PR #116, opened 2024-07-02
Frame has --json flag for JSON output support

Era-cell:


jviotti:
It's starting to shape up, but you should output the entire thing as one big JSON document, rather than parts of it as JSON. Then you wouldn't need to worry about indentation at all.

src/command_frame.cc
```diff
@@ -1,71 +1,98 @@
...
+                                   sourcemeta::jsontoolkit::default_schema_walker,
+                                   resolver(options))
+        .wait();
+
+    bool outputJson = options.contains("json") || options.contains("-j");
+
+    for (const auto &[key, entry] : frame) {
+        std::cout << "(LOCATION) URI: ";
+        std::cout << key.second << "\n";
+        std::cout << "    Schema           : " << entry.root.value_or("<ANONYMOUS>")
```
jviotti: i.e. all of these things you should be transforming to JSON. In the end you want the command to print only one big JSON document with all the information at once

suprith-hub: Ohh, I got it. I thought it might be this way, just in sometime...

src/command_frame.cc
```diff
@@ -1,71 +1,112 @@
...
-
-  for (const auto &[key, entry] : frame) {
-    std::cout << "(LOCATION) URI: ";
-    std::cout << key.second << "\n";
-    std::cout << "    Schema           : " << entry.root.value_or("<ANONYMOUS>")
-              << "\n";
-    std::cout << "    Pointer          :";
-    if (!entry.pointer.empty()) {
-      std::cout << " ";
+const char* enumToString(sourcemeta::jsontoolkit::ReferenceEntryType type) {
```
jviotti: You should never return a `const char *`. Please return an `std::string`

jviotti: Also, in C++, we don't use camelCase, but snake_case in variables and function calls. Please do `enum_to_string` or something like that. Same applies to various other parts in this PR

jviotti: Also, in modern C++, we use trailing return type style. i.e. `auto enum_to_string(...) -> result`

jviotti: Finally, this function doesn't modify its argument, so you want to make the argument `const`

jviotti: Also not strictly necessary in this case, but as a good practice, given this is a private function, please make it `static`

src/command_frame.cc
```diff
@@ -1,71 +1,112 @@
...
+    sourcemeta::jsontoolkit::frame(schema, frame, references,
+                                   sourcemeta::jsontoolkit::default_schema_walker,
+                                   resolver(options))
+        .wait();
 
-  for (const auto &[pointer, entry] : references) {
-    std::cout << "(REFERENCE) URI: ";
-    sourcemeta::jsontoolkit::stringify(pointer.second, std::cout);
-    std::cout << "\n";
+    bool outputJson = options.contains("json") || options.contains("j");
```
jviotti: All variables must be camel_case

jviotti: Also, try to use `const` whenever possible. This variable is not supposed to change, so it should be a constant

src/command_frame.cc
```diff
@@ -1,71 +1,112 @@
...
+    if (outputJson) {
+        sourcemeta::jsontoolkit::JSON outputJson = sourcemeta::jsontoolkit::JSON::make_object();
```
jviotti: You can do `auto` here

src/command_frame.cc
```diff
@@ -1,71 +1,112 @@
 #include <sourcemeta/jsontoolkit/json.h>
 #include <sourcemeta/jsontoolkit/jsonschema.h>
-
```
jviotti, line 3: Why was this line deleted?

docs/frame.markdown
```diff
@@ -58,3 +58,9 @@ reference:
 ```sh
 jsonschema frame path/to/my/schema.json
 ```
+
+### Frame a JSON Schema with output in the form of JSON
```
jviotti:
```suggestion
### Frame a JSON Schema and output result as a JSON document
```

jviotti:
CI is complaining about formatting. Try running `make` while having ClangFormat installed to re-format the code

jviotti:
Also, can you add a variant of this test (https://github.com/Intelligence-AI/jsonschema/blob/main/test/frame.sh) to exercise the new JSON functionality?

src/command_frame.cc
suprith-hub: Isn't specifying type always a good choice.. Why should we use auto in this case..?

src/command_frame.cc
suprith-hub: Cool! I got it

src/command_frame.cc
jviotti: You already are kinda specifying the type on the right hand side of assignment (`sourcemeta::jsontoolkit::JSON` from `sourcemeta::jsontookit::JSON::make_object`),  so this is a bit repetitive. That's why I suggested `auto`.

Era-cell:
@jviotti status says still  waiting for approval

jviotti:
@Era-cell Try now. I tweaked the project settings to be more permissive with first-time contributors

src/command_frame.cc
```diff
@@ -3,13 +3,26 @@
...
 #include "command.h"
 #include "utils.h"
 
-auto intelligence::jsonschema::cli::frame(
-    const std::span<const std::string> &arguments) -> int {
-  const auto options{parse_options(arguments, {})};
+static auto enum_to_string(const sourcemeta::jsontoolkit::ReferenceEntryType type) -> std::string {
+  switch (type) {
+  case sourcemeta::jsontoolkit::ReferenceEntryType::Resource:
+    return "Resource";
```
jviotti: Keep it all lowercase, like `resource` instead of `Resource`

test/frame.sh
```diff
@@ -70,4 +70,4 @@ cat << 'EOF' > "$TMP/expected.txt"
     - (fragment)     : /$defs/string
 EOF
 
-diff "$TMP/result.txt" "$TMP/expected.txt"
```
jviotti: This looks like an unintended change

test/frame_json_output.sh
```diff
@@ -0,0 +1,98 @@
...
+    "string": { "type": "string" }
+  }
+}
+EOF
+
+# Test with --json flag
+"$1" frame "$TMP/schema.json" --json > "$TMP/result_json.txt"
+
+# Test with -j flag
+"$1" frame "$TMP/schema.json" -j > "$TMP/result.json"
```
jviotti: I think you can just test with `--json`

src/command_frame.cc
```diff
@@ -21,51 +34,84 @@ auto intelligence::jsonschema::cli::frame(
...
 
-    sourcemeta::jsontoolkit::stringify(entry.pointer, std::cout);
-    std::cout << "\n";
-    std::cout << "    Base URI         : " << entry.base << "\n";
-    std::cout << "    Relative Pointer :";
-    if (!entry.relative_pointer.empty()) {
-      std::cout << " ";
+    for (const auto &[key, entry] : frame) {
+      auto frame_entry = sourcemeta::jsontoolkit::JSON::make_object();
+      frame_entry.assign("schema", sourcemeta::jsontoolkit::JSON{entry.root.value_or("<ANONYMOUS>")});
```
jviotti: If the root is not defined, can you make it null rather than setting it to a string with the value of `<ANONYMOUS>`?

test/frame_json_output.sh
```diff
@@ -0,0 +1,98 @@
...
+
+cat << 'EOF' > "$TMP/expected_json.txt"
+{
+  "frames": {
+    "https://example.com": {
+      "baseURI": "https://example.com",
+      "dialect": "https://json-schema.org/draft/2020-12/schema",
+      "pointer": "",
+      "relativePointer": "",
+      "schema": "https://example.com",
```
jviotti: Can you call these properties `root`, just to be consistent with the struct?

test/frame_json_output.sh
```diff
@@ -0,0 +1,98 @@
...
+"$1" frame "$TMP/schema.json" --json > "$TMP/result_json.txt"
+
+# Test with -j flag
+"$1" frame "$TMP/schema.json" -j > "$TMP/result.json"
+
+cat << 'EOF' > "$TMP/expected_json.txt"
+{
+  "frames": {
+    "https://example.com": {
+      "baseURI": "https://example.com",
```
jviotti: Here too. Maybe just call it `base`?

src/command_frame.cc
```diff
@@ -21,51 +34,84 @@ auto intelligence::jsonschema::cli::frame(
...
-      std::cout << "Dynamic";
-    } else {
-      std::cout << "Static";
+    for (const auto &[pointer, entry] : references) {
+      auto ref_entry = sourcemeta::jsontoolkit::JSON::make_object();
+      ref_entry.assign("type",
+                       sourcemeta::jsontoolkit::JSON{pointer.first == sourcemeta::jsontoolkit::ReferenceType::Dynamic ? "Dynamic" : "Static"});
+      ref_entry.assign("destination", sourcemeta::jsontoolkit::JSON{entry.destination});
+      if (entry.base.has_value()) {
+        ref_entry.assign("fragmentBaseURI", sourcemeta::jsontoolkit::JSON{entry.base.value()});
```
jviotti: Here you can make it just `base`. I think in general keeping the JSON properties consistent with the names in the struct will make it easier to maintain

src/command_frame.cc
suprith-hub: ookay... I did it because the source code had it Capitalized

test/frame_json_output.sh
Era-cell: For those whose pointer is empty(i.e, for first level) OR for all schema keywords: i.e,  `"schema": "https://example.com", `?

test/frame_json_output.sh
Era-cell: Instead of baseURI right?

test/frame.sh
suprith-hub: It doesn't have any difference between the original version, still I do not know why it says there is a diff

src/command_frame.cc
jviotti: Yeah, different conventions on different languages!

test/frame.sh
jviotti: It's the final new line character at the end of the file. We saw it in other PRs. It's likely a misconfiguration on your editor. Google how to make sure your editor includes an end-of-file line ending

jviotti: See the symbol that GitHub shows on the diff:

Screenshot 2024-07-11 at 11 54 41 PM

src/command_frame.cc
```diff
@@ -21,49 +36,96 @@ auto intelligence::jsonschema::cli::frame(
...
+    for (const auto &[pointer, entry] : references) {
+      auto ref_entry = sourcemeta::jsontoolkit::JSON::make_object();
+      ref_entry.assign("type", sourcemeta::jsontoolkit::JSON{
+                                   pointer.first ==
+                                           sourcemeta::jsontoolkit::ReferenceType::Dynamic
+                                       ? "Dynamic"
+                                       : "Static"});
+      ref_entry.assign("destination",
+                       sourcemeta::jsontoolkit::JSON{entry.destination});
+      if (entry.base.has_value()) {
```
jviotti, line 88: For every `has_value` check, can you add an `else` that sets it to null? Just to make it explicit that such property doesn't have a value

test/frame.sh
suprith-hub: Shet! sorry I got it, I think myself I removed the EOF line, I will correct it

Era-cell:
@jviotti what is the clang formatter I should be using? I used llvm clang with [vscode extention](https://github.com/xaverh/vscode-clang-format), but its still failing

jviotti:
@Era-cell I don't use VS Code, so not sure about that plugin, but you might be installing a different version? Here is how we do it for all platform on GitHub Actions: https://github.com/Intelligence-AI/jsonschema/blob/main/.github/workflows/test.yml#L40-L41

Era-cell:
> @Era-cell I don't use VS Code, so not sure about that plugin, but you might be installing a different version? Here is how we do it for all platform on GitHub Actions: https://github.com/Intelligence-AI/jsonschema/blob/main/.github/workflows/test.yml#L40-L41

@jviotti I did format using clang-format version 18.1.5 Python, it still fails, idk why

jviotti:
Do you get an error, or do you mean the formatting is not happening how you would expect it to happen?

Era-cell:
@jviotti looks like clang formatting worked for windows but not for mac, and ubuntu

jviotti:
Formatting is OK now, it's the test that's failing. We don't run tests on Windows (yet) so that's why that's green.

Era-cell:
Ok, its good.. but can you check the changes...

test/frame/pass_json_output.sh
```diff
@@ -0,0 +1,101 @@
...
+      "base": "https://example.com",
+      "destination": "https://example.com#/$defs/string",
+      "fragment": "/$defs/string",
+      "type": "Static"
+    },
+    "/$schema": {
+      "base": "https://json-schema.org/draft/2020-12/schema",
+      "destination": "https://json-schema.org/draft/2020-12/schema",
+      "fragment": null,
+      "type": "Static"
```
jviotti: Can you make these lowercase too, like the other types? i.e. `static` and `dynamic`

test/frame/pass.sh
jviotti, line 1: Thanks a lot for correcting these too 🙏🏻

jviotti:
It's looking really good! Just one minor comment and I think we are done

src/command_frame.cc
```diff
@@ -27,75 +42,140 @@ auto intelligence::jsonschema::cli::frame(
...
+    output_json_object.assign("frames",
+                              sourcemeta::jsontoolkit::JSON{frame_json});
+
+    for (const auto &[pointer, entry] : references) {
+      auto ref_entry = sourcemeta::jsontoolkit::JSON::make_object();
+      ref_entry.assign(
+          "type",
+          sourcemeta::jsontoolkit::JSON{
+              pointer.first == sourcemeta::jsontoolkit::ReferenceType::Dynamic
+                  ? "Dynamic"
```
jviotti:
```suggestion
                  ? "dynamic"
```

src/command_frame.cc
```diff
@@ -27,75 +42,140 @@ auto intelligence::jsonschema::cli::frame(
...
+                  : "Static"});
```
jviotti:
```suggestion
                  : "static"});
```

Era-cell:
@jviotti  I have updated the PR from my side.. can you check it, if I have missed any other case

jviotti:
Thanks a lot! Great work, and I hope you enjoyed C++ :D I'll release a new version with this change soon and announce it on LinkedIn

Era-cell:
> Thanks a lot! Great work, and I hope you enjoyed C++ :D I'll release a new version with this change soon and announce it on LinkedIn

Thank you for guiding me 😁, I took a lot of time 😅...

PR #197, opened 2024-11-08
feat:Make use of coloured output

Pavankumar07s:
Task 1 completed:Create a first PR that introduce the <unistd.h> approach on a helper in utils.h/utils.cc and maybe just uses it to color code "JSON Schema CLI vX.Y.Z" on the help page. Something super simple and without letting users control it yet.
Screenshot from 2024-11-08 21-36-05


src/messages.h
```diff
@@ -0,0 +1,14 @@
+#ifndef MESSAGES_H
```
jviotti: Please do if guards in a consistent manner, like the other headers. For example, `SOURCEMETA_JSONSCHEMA_CLI_MESSAGES_H_`

src/main.cc
```diff
@@ -8,9 +8,16 @@
 #include <string>      // std::string
 #include <string_view> // std::string_view
 #include <vector>      // std::vector
+#ifdef _WIN32
+#include <io.h>
+#define isatty _isatty
+#else
```
jviotti, line 14: Where are we actually checking for TTY in the code?

src/main.cc
```diff
@@ -85,101 +92,157 @@ Global Options:
 
 For more documentation, visit https://github.com/sourcemeta/jsonschema
 )EOF"};
+// Initialize color usage
+bool use_colors = true;
```
jviotti: We shouldn't have a boolean like this. Instead, each command should check the command line arguments + the TTY check.

src/main.cc
```diff
@@ -85,101 +92,157 @@ Global Options:
...
 
 auto jsonschema_main(const std::string &program, const std::string &command,
-                     const std::span<const std::string> &arguments) -> int {
-  if (command == "fmt") {
+                     const std::span<const std::string> &arguments) -> int
```
jviotti: All of these changes seem unrelated? They will also break the build, and we check formatting against `ClangFormat`

src/main.cc
```diff
@@ -85,101 +92,157 @@ Global Options:
...
+
+    print_success(std::string(USAGE_DETAILS));
+
     return EXIT_SUCCESS;
   }
 }
 
-auto main(int argc, char *argv[]) noexcept -> int {
-  try {
+auto main(int argc, char *argv[]) noexcept -> int
```
jviotti: Most of these seem to be unrelated formatting changes?

src/messages.h
```diff
@@ -0,0 +1,14 @@
...
+#define MESSAGES_H
+
+#include <iostream>
+#include <string>
+#include "../include/termcolor/termcolor.hpp"
+
+// Declare color control as an external variable
+extern bool use_colors;
```
jviotti: We should avoid these kind of global variables, even more `extern` ones.

src/messages.cc
jviotti, line 1: Mainly for this PR, I don't think we need these extra helpers. Can you use `termcolor` on standard streams directly?

include/termcolor/termcolor.hpp
jviotti, line 1: You shouldn't be including untracked external dependencies. Please add them using `DEPENDENCIES` and https://github.com/sourcemeta/vendorpull

jviotti: And once you add it like that, you will need to pull the dependency using CMake's `find_package`, like we do for other dependencies

jviotti (changes requested):
Left a few comments

src/main.cc
```diff
@@ -85,101 +92,157 @@ Global Options:
...
-              << " <command> [arguments...]\n";
-    std::cout << USAGE_DETAILS;
+  }
+  else
+  {
+    print_success(std::string("JSON Schema CLI - v") + std::string(sourcemeta::jsonschema::cli::PROJECT_VERSION));
+
+    print_success("Usage: " + std::filesystem::path{program}.filename().string() + " <command> [arguments...]");
+
+    print_success(std::string(USAGE_DETAILS));
```
jviotti: We probably shouldn't colorise the entire thing. Maybe just colorise the usage line above or something like that?

src/main.cc
Pavankumar07s: Yes

Pavankumar07s: We haven't introduced yet . In this pr i have just inlmported the module for tty and color. In next pr we will introduce condition for tty as cli-program detected then we will continue using colors else we will turn off.

src/main.cc
Pavankumar07s: Alright ,in our code i was just going to check ones when the problem starts.Thankyou for guidance i will make sure it checks each time when command run.

src/main.cc
Pavankumar07s: Thankyou ,i was also confused i will now use formated code for our simplicity and build.

src/main.cc
Pavankumar07s: I am going to revert al these formatting. Thankyou.

src/messages.h
Pavankumar07s: Alright @jviotti 👍

src/messages.cc
Pavankumar07s: Yes we can do but in further steps ,we again have to change .Is is better to define all colors in one messages page so that it would be easy and maintained. Pls share your thoughts.

src/main.cc
Pavankumar07s: Lets just color the usage part.

include/termcolor/termcolor.hpp
Pavankumar07s: yeaah i got it .Thankyou.

src/messages.cc
jviotti: I think `termcolor` interface is already quite concise and we could use it directly when needed.

src/main.cc
jviotti: If you run `make` on your system that should run `ClangFormat` as per the project's configuration and always get the formatting right

src/main.cc
jviotti: I think you still need the TTY check on this one, otherwise i.e. piping the help text will result in escaped characters

src/messages.cc
Pavankumar07s: Alright @jviotti 👍

src/main.cc
Pavankumar07s: Yeah i am this one only Thankyou you are great help.😃

src/main.cc
Pavankumar07s: Alright @jviotti 👍

Pavankumar07s:
Hiii @jviotti whats i am doing wrong ? When i am using make command it is showing 100% test passes but the workflow is breaking...can you guide me.
Screenshot from 2024-11-13 13-13-32
Screenshot from 2024-11-13 12-54-46
Screenshot from 2024-11-13 12-55-05

PR #199, opened 2024-11-13
feat:add color and ttl check

Pavankumar07s:
checked all the points.
Got 100perset test passed.
Screenshot from 2024-11-14 00-11-50

Screenshot from 2024-11-14 00-11-37

Screenshot from 2024-11-14 00-14-28


Pavankumar07s:
@jviotti, you're a great help. But can you assist me with this broken test?

Pavankumar07s:
Alright i think i got an error.

Pavankumar07s:
hiii @jviotti can you assist me for windows workflow.

CMakeLists.txt
```diff
@@ -2,7 +2,7 @@ cmake_minimum_required(VERSION 3.16)
 project(jsonschema VERSION 4.3.0 LANGUAGES CXX)
 list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake")
 include(vendor/noa/cmake/noa.cmake)
-
+include_directories(vendor/termcolor/include)
```
jviotti, line 5: Note `termcolor` is a CMake project with its own `CMakeLists.txt`, so instead of including its files directly here (which is not a best practice), you should create a `cmake/FindTermColor.cmake` and them here do `find_package(TermColor REQUIRED)` like we do for other dependencies

src/main.cc
```diff
@@ -12,6 +12,17 @@
 #include "command.h"
 #include "configure.h"
 
+#ifdef _WIN32
+#include <io.h>
+#define isatty _isatty
+
```
jviotti, line 18: Unnecessary blank line?

src/main.cc
```diff
@@ -12,6 +12,17 @@
...
+#define fileno _fileno
+#else
+
```
jviotti, line 21: Unnecessary blank line?

src/main.cc
Pavankumar07s: Alright 👍 i will remove it.

src/main.cc
```diff
@@ -88,6 +99,13 @@ For more documentation, visit https://github.com/sourcemeta/jsonschema
 
 auto jsonschema_main(const std::string &program, const std::string &command,
                      const std::span<const std::string> &arguments) -> int {
+  bool use_colors = true;
+  if (std::find(arguments.begin(), arguments.end(), "--no-color") !=
```
jviotti, line 103: We have a `parse_options` helper function to avoid re-parsing command line arguments in every place. Can you make use of that instead?

Also, you probably want to support a short version of this. Maybe `-n`?

Finally, you should document this new option on the help page and on `docs/`

src/main.cc
Pavankumar07s: Alright i will also remove it.

src/main.cc
```diff
@@ -117,7 +135,12 @@ auto jsonschema_main(const std::string &program, const std::string &command,
               << sourcemeta::jsonschema::cli::PROJECT_VERSION << "\n";
     std::cout << "Usage: " << std::filesystem::path{program}.filename().string()
               << " <command> [arguments...]\n";
-    std::cout << USAGE_DETAILS;
+    if (use_colors) {
+      std::cout << termcolor::yellow << USAGE_DETAILS << termcolor::reset
```
jviotti, line 139: I don't think it makes much sense to color out the entire help. Maybe just color out the project name in line 134?

CMakeLists.txt
Pavankumar07s: Alright @jviotti got your point.

Pavankumar07s:
@jviotti you tell me why windows build is getting failed?

vendor/termcolor/.mailmap
jviotti, line 1: Can you mask all unnecessary files (except `LICENSE`) using `vendorpull`? See https://github.com/sourcemeta/vendorpull#masking. Otherwise we are even including screenshots from termcolor here

jviotti:
I'm not sure. The error is weird indeed. Let's fix all of the other things, and if it still fails when the PR is ready to merge, I'll take a proper look!

src/main.cc
Pavankumar07s: Alright ,i am getting really exited.

Pavankumar07s:
Hii @jviotti, I know this is a bit of an unusual request, but I was wondering if you could share some resources that could help me delve deeper into this topic.

jviotti:
Hey @Pavankumar07s, about what topic specifically? Maybe this page where we list useful resources in general help? https://www.sourcemeta.com/resources/

src/main.cc
jviotti: Actually, just to keep things simpler and merge this sooner, let's avoid `--no-color` altogether. Just do it purely depending on the TTY stuff. We can then get into parsing that option on specific commands later on

Pavankumar07s: I am not feeling good,i think my health is not good.I am going home i will work on this in few days. Thanks.

Pavankumar07s:
I am not feeling good,i think my health is not good.I am going home i will work on this in few days. Thanks.

Pavankumar07s:
I have already shared my situation, but I think it was in an unnoticed place.

PR #220, opened 2025-02-02
Feat: Add --path Option to Target Specific Schema Parts in validate Command

### LINKED ISSUE #218: Add a `--path` option to commands like `validate`
That option would take a JSON Pointer to say: "act on the schema located in this part of the input". It can be a low-hanging fruit to start allowing various of these commands to operate on OpenAPI spec.

Karan-Palan:
**What kind of change does this PR introduce?**

- Feature: Added support for the `--path` (or `-p`) option in the `validate` command to allow targeting specific parts of a JSON schema using JSON Pointers.

**Issue Number:**

- Closes #218 

**Screenshots/videos:**

[Screencast from 2025-02-02 12-53-07.webm](https://github.com/user-attachments/assets/f6dc7ece-0018-42e6-88cb-5536892afb8c)

**If relevant, did you update the documentation?**

Not yet

**Summary**

This PR introduces the `--path` (`-p`) option to the `validate` command, enabling users to specify a JSON Pointer for validating a sub-schema within a larger JSON schema. This is especially useful for OpenAPI specs and complex JSON structures where validation may be needed for specific nested sections.

Key changes:
- Added `--path`/`-p` option to CLI parser.
- Integrated JSON Pointer resolution to extract the specified sub-schema.
- Added error handling for invalid pointers or sub-schemas.

**Does this PR introduce a breaking change?**

- No breaking changes. The new option is fully backward-compatible. Existing workflows without the `--path` option will continue to function as expected.


src/command_validate.cc
```diff
@@ -20,7 +25,8 @@
 auto sourcemeta::jsonschema::cli::validate(
     const std::span<const std::string> &arguments) -> int {
   const auto options{
-      parse_options(arguments, {"h", "http", "b", "benchmark", "t", "trace"})};
+      parse_options(arguments, {"h", "http", "b", "benchmark", "t", "trace",
+                               "p", "path"})};
```
jviotti: This `parse_options` function is only for flags that don't take arguments. I don't think you need to add it here!

jviotti:
Awesome! Great first step. Can you fix the formatting issues and add a couple of test cases (at least one where the pointer exists and one where it doesn't)?

src/command_validate.cc
Karan-Palan: I manually parsed --path (or -p) from the leftover positional argument

Karan-Palan:
> Awesome! Great first step. Can you fix the formatting issues and add a couple of test cases (at least one where the pointer exists and one where it doesn't)?

Thanks!
I fixed the linting errors - they were because my clang default was set to chromium's source. I also added some tests and docs (it's my first time writing tests in shell :p)
Please review when you get the time!

jviotti:
@Karan-Palan I think you forgot to actually register the tests, so they are not really running right now. You can do so like this:

```diff
diff --git a/test/CMakeLists.txt b/test/CMakeLists.txt
index 0f9840b..a3baa6d 100644
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -74,6 +74,8 @@ add_jsonschema_test_unix(validate/pass_many_verbose)
 add_jsonschema_test_unix(validate/fail_many)
 add_jsonschema_test_unix(validate/fail_many_verbose)
 add_jsonschema_test_unix(validate/fail_yaml)
+add_jsonschema_test_unix(validate/pass_pointer_exists)
+add_jsonschema_test_unix(validate/fail_pointer_not_exists)

 # Metaschema
 add_jsonschema_test_unix(metaschema/pass_trace)
```

With that, running `make` does reveal a test failure!

docs/validate.markdown
```diff
@@ -85,6 +85,12 @@ jsonschema validate path/to/my/schema.json path/to/my/instance.json \
   --resolve path/to/external.json
 ```
 
+### Validate a JSON instance against a sub-schema
```
jviotti:
```suggestion
### Validate a JSON instance against a specific part of the schema given a JSON Pointer
```

Just to make it more explicit?

docs/validate.markdown
jviotti, line 1: Can you mention `--path` in the command signature at the top of this file? And also in the help section in `main.cc`

src/command_validate.cc
```diff
@@ -22,48 +27,97 @@ auto sourcemeta::jsonschema::cli::validate(
...
+  // Attempt to extract a pointer value from leftover arguments:
+  //    --path=<pointer>
+  //    --path <pointer>
+  //    -p <pointer>
+  std::optional<std::string> pointer_option;
+  for (auto it = leftover.begin(); it != leftover.end();) {
+    const auto &arg = *it;
+
+    // a) --path=/something
+    if (arg.rfind("--path=", 0) == 0) {
```
jviotti: I believe `options.at("p")` and `options.at("path")` will give you what you want? Look at how we take options in other commands, like for `--resolve`

test/validate/pass_pointer_exists.sh
```diff
@@ -0,0 +1,43 @@
...
+cat << 'EOF' > "$TMP/instance.json"
+{
+  "foo": 123
+}
+EOF
+
+if ! "$1" validate \
+    "$TMP/schema.json" \
+    "$TMP/instance.json" \
+    --path=/properties/nested \
```
jviotti: Instead of accessing a subschema of the schema, to make the test more representative, can you make the schema a very concise OpenAPI spec (in JSON) and access one of its schema components?

test/validate/pass_pointer_exists.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/sh
+
+set -o errexit
+set -o nounset
+
+TMP="$(mktemp -d)"
+cleanup() { rm -rf "$TMP"; }
```
jviotti: Minor thing, but other commands call this prelude `clean`. Can you do the same for consistency?

test/validate/pass_pointer_exists.sh
```diff
@@ -0,0 +1,43 @@
...
+}
+EOF
+
+cat << 'EOF' > "$TMP/instance.json"
+{
+  "foo": 123
+}
+EOF
+
+if ! "$1" validate \
```
jviotti: Also, you don't need the explicit `if` check. Because of `set -o errexit`, the entire script will fail if any of its commands fail. See how we do it for other "pass" tests

test/validate/pass_pointer_exists.sh
```diff
@@ -0,0 +1,43 @@
...
+    "$TMP/instance.json" \
+    --path=/properties/nested \
+    1>/dev/null 2>"$TMP/stderr.log"
+then
+  echo "Expected success, got failure!"
+  cat "$TMP/stderr.log"
+  exit 1
+fi
+
+echo "pass_pointer_exists.sh: PASS"
```
jviotti: No need to print these things. `ctest` will already nicely report the specific test cases

test/validate/fail_pointer_not_exists.sh
jviotti, line 1: Can you also add a test where the JSON Pointer passed to `--path` is indeed valid, but it's not an actual schema? Maybe `/type` or something like that?

docs/validate.markdown
Karan-Palan: Sure

PR #221, opened 2025-02-02
Feat:Termcolor

### LINKED ISSUE #1: Make use of coloured output
See https://github.com/ikalnytskyi/termcolor

Pavankumar07s:
What kind of change does this PR introduce?

Feature: 1.Added  termcolor
                2. just colored the version 
                3.used parse_options helper function to avoid re-parsing command line arguments.
                4. added short version -n for --no-color
Issue Number:

Closes #1 
Screenshots/videos:
Screenshot from 2025-02-02 15-28-27
Screenshot from 2025-02-02 15-28-40
Screenshot from 2025-02-02 15-28-51


Pavankumar07s:
Hello Mr. @jviotti, I am using clang-format to format the code, but the workflow still fails. Could you please guide me?

jviotti:
Hey @Pavankumar07s, does this PR take over https://github.com/sourcemeta/jsonschema/pull/199? Can we close that one if so?

> Hello Mr. @jviotti, I am using clang-format to format the code, but the workflow still fails. Could you please guide me?

How are you installing and running `clang-format`? If you have a too-old version, it might be doing things differently. See https://github.com/sourcemeta/jsonschema/blob/main/.github/workflows/test.yml#L41 for the version we use on CI at any given point in time.

Also note that we run `clang-format` with a specific configuration file (https://github.com/sourcemeta/core/blob/main/cmake/common/targets/clang-format.config) and if you are using a code editor integration, it might not be picking it up.

I always run it with `make` (or `nmake` on Windows). That builds the project and runs the formatter with all the right settings.

cmake/FindTermColor.cmake
```diff
@@ -0,0 +1,13 @@
...
+endif()
+
+add_library(termcolor INTERFACE)
+add_library(termcolor::termcolor ALIAS termcolor)
+
+target_include_directories(termcolor INTERFACE
+    ${CMAKE_CURRENT_LIST_DIR}/../vendor/termcolor/include)
+
+include(FindPackageHandleStandardArgs)
+find_package_handle_standard_args(TermColor
```
jviotti, line 12: What is this for?

vendor/termcolor/.mailmap
jviotti, line 1: Can you mask all of the files that are not essential for the build? See https://github.com/sourcemeta/vendorpull?tab=readme-ov-file#masking

jviotti:
Looking better. Can you also implement a couple of tests cases showing the color escaping codes taking effect, and also the "no color" option doing the right thing?

Pavankumar07s:
Alright Mr @jviotti

cmake/FindTermColor.cmake
Pavankumar07s: Mr. @jviotti, as you suggested, termcolor has its own CMake file, and I need to create this file to use it. I researched and came up with this approach.

vendor/termcolor/.mailmap
Pavankumar07s: Alright Mr @jviotti

cmake/FindTermColor.cmake
jviotti, line 1: This is overly complicated and not following the same conventions we have in other `Find*` files. Plus it's using the `_FOUND` suffixes that CMake actually expects, relying on targets instead. This works for me and it is pretty much exactly the same as with other dependencies:

```cmake
if(NOT TermColor_FOUND)
  add_subdirectory("${PROJECT_SOURCE_DIR}/vendor/termcolor")
  set(TermColor_FOUND ON)
endif()
```

cmake/FindTermColor.cmake
jviotti: See https://github.com/sourcemeta/jsonschema/pull/221/files#r1941593287. I think you are doing it the hard way

src/command_validate.cc
jviotti, line 1: This seems like an unrelated change?

vendor/termcolor/.mailmap
jviotti, line 1: Let's mask all of these files!

jviotti:
Maybe to simplify things, you can first send a PR integrating `TermColor` and linking to it? We can get that merged, and then focus on the actual option?

vendor/termcolor/.mailmap
Pavankumar07s: Ohhk sir , reading the docs.

cmake/FindTermColor.cmake
Pavankumar07s: Yes changing to this way.

Issue #58, opened 2024-06-05
Output `lint` and `validate` results in JSON if `--json` is passed


jviotti:

Karan-Palan:
I'll work on `validate` soon!

jviotti:
It took me a while, but here is a PR introducing standard output format support for Blaze: https://github.com/sourcemeta/blaze/pull/471

PR #228, opened 2025-02-12
feat: add --json to lint command

Karan-Palan:
**What kind of change does this PR introduce?**

- Feature: Added support for the `--json` (or `-j`) option in the `lint` command.

**Issue Number:**

-  #58 

**Screenshots/videos:**

image

**If relevant, did you update the documentation?**

Not yet

### DIFF
--- a/src/command_lint.cc
+++ b/src/command_lint.cc
@@ -3,16 +3,18 @@
 #include <sourcemeta/core/jsonpointer.h>
 #include <sourcemeta/core/jsonschema.h>
 
-#include <cstdlib>  // EXIT_SUCCESS
+#include <cstdlib>  // EXIT_SUCCESS, EXIT_FAILURE
 #include <fstream>  // std::ofstream
 #include <iostream> // std::cerr, std::cout
+#include <sstream>  // std::ostringstream
 
 #include "command.h"
 #include "utils.h"
 
 auto sourcemeta::jsonschema::cli::lint(
     const std::span<const std::string> &arguments) -> int {
-  const auto options{parse_options(arguments, {"f", "fix"})};
+  const auto options{parse_options(arguments, {"f", "fix", "json", "j"})};
+  const bool output_json = options.contains("json") || options.contains("j");
 
   sourcemeta::core::SchemaTransformer bundle;
   sourcemeta::core::add(bundle,
@@ -47,7 +49,16 @@ auto sourcemeta::jsonschema::cli::lint(
                                  sourcemeta::core::schema_format_compare);
       output << "\n";
     }
+
+    // If --fix and --json were both requested, optionally print a small JSON:
+    if (output_json) {
+      auto msg = sourcemeta::core::JSON::make_object();
+      msg.assign("fixApplied", sourcemeta::core::JSON{true});
+      sourcemeta::core::prettify(msg, std::cout);
+      std::cout << std::endl;
+    }
   } else {
+    auto issues_array = sourcemeta::core::JSON::make_array();
     for (const auto &entry :
          for_each_json(options.at(""), parse_ignore(options),
                        parse_extensions(options))) {
@@ -55,12 +66,29 @@ auto sourcemeta::jsonschema::cli::lint(
       const bool subresult = bundle.check(
           entry.second, sourcemeta::core::schema_official_walker,
           resolver(options),
-          [&entry](const auto &pointer, const auto &name, const auto &message) {
-            std::cout << entry.first.string() << ":\n";
-            std::cout << "  " << message << " (" << name << ")\n";
-            std::cout << "    at schema location \"";
-            sourcemeta::core::stringify(pointer, std::cout);
-            std::cout << "\"\n";
+          [&](const auto &pointer, const auto &name, const auto &message) {
+            if (output_json) {
+              // Collect in a JSON object instead of printing lines
+              auto error_obj = sourcemeta::core::JSON::make_object();
+
+              error_obj.assign("file",
+                               sourcemeta::core::JSON{entry.first.string()});
+              error_obj.assign("rule", sourcemeta::core::JSON{name});
+              error_obj.assign("message", sourcemeta::core::JSON{message});
+
+              std::ostringstream pointer_stream;
+              sourcemeta::core::stringify(pointer, pointer_stream);
+              error_obj.assign("pointer",
+                               sourcemeta::core::JSON{pointer_stream.str()});
+
+              issues_array.push_back(error_obj);
+            } else {
+              std::cout << entry.first.string() << ":\n";
+              std::cout << "  " << message << " (" << name << ")\n";
+              std::cout << "    at schema location \"";
+              sourcemeta::core::stringify(pointer, std::cout);
+              std::cout << "\"\n";
+            }
           });
 
       if (subresult) {
@@ -68,6 +96,13 @@ auto sourcemeta::jsonschema::cli::lint(
       } else {
         result = false;
       }
+    }
+    if (output_json) {
+      auto output_json_object = sourcemeta::core::JSON::make_object();
+      output_json_object.assign("passed", sourcemeta::core::JSON{result});
+      output_json_object.assign("issues", sourcemeta::core::JSON{issues_array});
+      sourcemeta::core::prettify(output_json_object, std::cout);
+      std::cout << std::endl;
     }
   }
 


Karan-Palan:
Based on this [discussion](https://json-schema.slack.com/archives/C8C4UBXDF/p1739193220641959?thread_ts=1738590781.293879&cid=C8C4UBXDF). @jviotti PTAL, if it looks fine, I'll add some tests and update the docs.

jviotti:
@Karan-Palan Great start! Some minor tweaks to be more JSON Schema idiomatic:

- Can you rename `passed` to `valid`?
- Can you rename `issues` to `errors`?
- Can you rename `file` to `path`?
- Can you rename `rule` to `id`?
- Can you rename `pointer` to `schemaLocation`?

Overall, just trying to make it match JSON Schema Standard Output Formats a bit, at least in style!

src/command_lint.cc
```diff
@@ -47,20 +49,46 @@ auto sourcemeta::jsonschema::cli::lint(
                                  sourcemeta::core::schema_format_compare);
       output << "\n";
     }
+
+    // If --fix and --json were both requested, optionally print a small JSON:
```
jviotti: In similar cases, other linters will output exactly the same, just that if you run it again, there will be no issues. Let's adopt that, as developers will be familiar with that

jviotti:
I also left a tiny comment, but looks pretty good. Looking forward to some tests and the docs

Karan-Palan:
> I also left a tiny comment, but looks pretty good. Looking forward to some tests and the docs

Done, PTAL

docs/lint.markdown
```diff
@@ -27,6 +27,8 @@ automatically fix many of them.
 
 **The `--fix/-f` option is not supported when passing YAML schemas.**
 
+**The `--json/-j` option prints results as a JSON object rather than human‐readable text.**
```
jviotti: I would remove this line. We don't have it for `inspect` (which also supports the same flag) and I guess it is pretty self explanatory

test/lint/fail_lint_enum_with_type.sh
jviotti, line 1: This test is not exercising the `--json` flag? We have linting tests already. The point is to do linting with `--json` on the new ones!

test/lint/pass_lint_fix_single_enum.sh
jviotti, line 1: Same here. This test is not exercising the `--json` flag.

src/command_lint.cc
```diff
@@ -47,20 +49,44 @@ auto sourcemeta::jsonschema::cli::lint(
                                  sourcemeta::core::schema_format_compare);
       output << "\n";
     }
+
+    if (output_json) {
+      auto msg = sourcemeta::core::JSON::make_object();
+      msg.assign("fixApplied", sourcemeta::core::JSON{true});
```
jviotti: We don't want this. If `--fix` was passed, still print the normal JSON output

src/main.cc
```diff
@@ -45,11 +45,12 @@ Global Options:
        Format the input schemas in-place or check they are formatted.
        This command does not support YAML schemas yet.
 
-   lint [schemas-or-directories...] [--fix/-f] [--extension/-e <extension>]
+   lint [schemas-or-directories...] [--fix/-f] [--json/-j] [--extension/-e <extension>]
         [--ignore/-i <schemas-or-directories>]
 
        Lint the input schemas and potentially fix the reported issues.
        The --fix/-f option is not supported when passing YAML schemas.
+       Use --json/-j to output lint errors and warnings in JSON format.
```
jviotti:
```suggestion
       Use --json/-j to output lint errors in JSON.
```

jviotti:
Looking good! Can you add one more tests exercising `--fix` alongside `--json`?

Karan-Palan:
Done, ptal

test/lint/pass_lint_fix_json.sh
```diff
@@ -0,0 +1,39 @@
...
+EOF
+
+"$1" lint "$TMP/schema.json" --fix --json >"$TMP/output.json" 2>&1 && CODE="$?" || CODE="$?"
+test "$CODE" = "0" || exit 1
+
+cat << EOF > "$TMP/expected.json"
+{
+  "valid": true,
+  "errors": [],
+  "fixApplied": true
```
jviotti: Let's get rid of this flag. We shouldn't report it. Usually what linters do with `--fix` is that they will still report ALL the errors that were fixed (as if you didn't pass `--fix`) but output code 0. And if you run it again, you get no errors reported at all

test/lint/pass_lint_json.sh
```diff
@@ -0,0 +1,26 @@
...
+
+cat << 'EOF' > "$TMP/schema.json"
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "const": "foo"
+}
+EOF
+
+"$1" lint "$TMP/schema.json" --json >"$TMP/output.json" 2>&1 && CODE="$?" || CODE="$?"
+test "$CODE" = "0" || exit 1
```
jviotti:
```suggestion
test "$CODE" = "0"
```

You don't need the OR. The entire script should fail on any line failures

test/lint/pass_lint_json.sh
```diff
@@ -0,0 +1,26 @@
...
+set -o nounset
+
+TMP="$(mktemp -d)"
+clean() { rm -rf "$TMP"; }
+trap clean EXIT
+
+cat << 'EOF' > "$TMP/schema.json"
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "const": "foo"
```
jviotti: This schema is invalid. `const` is not present in Draft 4. That's a linter rule we should have in the future I guess. For now, can you make sure this schema is at least valid? i.e.

```suggestion
  "enum": [ "foo" ]
```

test/lint/pass_lint_json.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/sh
+set -o errexit
```
jviotti, line 3: Can you leave a blank line after the hash bang, like other scripts? Same in the other ones you added here

src/command_lint.cc
```diff
@@ -47,29 +51,53 @@ auto sourcemeta::jsonschema::cli::lint(
...
+  if (output_json) {
+    auto output_json_object = sourcemeta::core::JSON::make_object();
+    output_json_object.assign("valid", sourcemeta::core::JSON{result});
+    output_json_object.assign("errors", sourcemeta::core::JSON{errors_array});
+    if (fix) {
+      output_json_object.assign("fixApplied", sourcemeta::core::JSON{true});
+    }
+    sourcemeta::core::prettify(output_json_object, std::cout);
+    std::cout << "\n";
+  } else {
```
jviotti: No need to have an `else` clause that does nothing. Can you delete it?

src/command_lint.cc
```diff
@@ -47,29 +51,53 @@ auto sourcemeta::jsonschema::cli::lint(
...
+    if (!subresult) {
+      result = false;
     }
   }
+  if (output_json) {
+    auto output_json_object = sourcemeta::core::JSON::make_object();
+    output_json_object.assign("valid", sourcemeta::core::JSON{result});
+    output_json_object.assign("errors", sourcemeta::core::JSON{errors_array});
+    if (fix) {
+      output_json_object.assign("fixApplied", sourcemeta::core::JSON{true});
```
jviotti: See my comment below. We shouldn't be reporting this

test/lint/pass_lint_json.sh
Karan-Palan: Just added those in my rules :D

Karan-Palan:
Done. Apologies for the extra commits for adding blanks after hashes, my formatter for bash files removes extra lines (below hash and last line)

test/lint/pass_lint_fix_json.sh
```diff
@@ -0,0 +1,39 @@
...
+cat << 'EOF' > "$TMP/schema.json"
+{
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "type": "string",
+  "enum": [ "foo" ]
+}
+EOF
+
+"$1" lint "$TMP/schema.json" --fix --json >"$TMP/output.json" 2>&1 && CODE="$?" || CODE="$?"
+test "$CODE" = "0" || exit 1
```
jviotti: You don't need to test code exit code 0. If you don't put any of this, non-zero codes will fail out of the box. Let me give you some suggested changes

```suggestion
```

test/lint/pass_lint_fix_json.sh
```diff
@@ -0,0 +1,39 @@
...
+
+cat << 'EOF' > "$TMP/schema.json"
+{
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "type": "string",
+  "enum": [ "foo" ]
+}
+EOF
+
+"$1" lint "$TMP/schema.json" --fix --json >"$TMP/output.json" 2>&1 && CODE="$?" || CODE="$?"
```
jviotti:
```suggestion
"$1" lint "$TMP/schema.json" --fix --json >"$TMP/output.json" 2>&1
```

test/lint/pass_lint_fix_json.sh
```diff
@@ -0,0 +1,39 @@
...
+cat << 'EOF' > "$TMP/expected_fixed.json"
+{
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "const": "foo"
+}
+EOF
+
+diff "$TMP/schema.json" "$TMP/expected_fixed.json"
+
+echo "pass_lint_fix_json: PASS"
```
jviotti:
```suggestion
```

No need for this!

jviotti:
I left a few minor remaining suggestions, but its almost there!

jviotti:
Actually, I guess I can just accept them myself and just merge. Let me see

test/lint/pass_lint_fix_json.sh
```diff
@@ -0,0 +1,39 @@
...
+
+cat << 'EOF' > "$TMP/expected_fixed.json"
+{
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "const": "foo"
+}
+EOF
+
+diff "$TMP/schema.json" "$TMP/expected_fixed.json"
+
```
jviotti:
```suggestion
```

test/lint/pass_lint_json.sh
```diff
@@ -0,0 +1,27 @@
...
+trap clean EXIT
+
+cat << 'EOF' > "$TMP/schema.json"
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "enum": [ "foo" ]
+}
+EOF
+
+"$1" lint "$TMP/schema.json" --json >"$TMP/output.json" 2>&1 && CODE="$?" || CODE="$?"
```
jviotti:
```suggestion
"$1" lint "$TMP/schema.json" --json >"$TMP/output.json" 2>&1
```

test/lint/pass_lint_json.sh
```diff
@@ -0,0 +1,27 @@
...
+test "$CODE" = "0"
```
jviotti:
```suggestion
```

jviotti (approved):

Karan-Palan:
Thanks for your patience, @jviotti i! I really appreciate the feedback. I'll ensure that my future test cases follow the same format.

jviotti:
Thank a lot for the work! I'll release a new version of the CLI in a bit

jviotti:
Done! See v7.1.0

Issue #249, opened 2025-03-06
cannot change permissions of ‘/usr/local/bin’ since 2025-03-02


v1v:
Start failing with 
```
install: cannot change permissions of ‘/usr/local/bin’: Operation not permitted
```

```
Run sourcemeta/jsonschema@2470d6ca67dc844b7048ee285e96b7652723756a
Run curl --retry 5 --location --fail-early --silent --show-error \
---- Fetching the pre-built JSON Schema CLI binary from GitHub Releases
---- Fetching version v6.0.3 from https://github.com/sourcemeta/jsonschema/releases/download/v6.0.3/jsonschema-6.0.3-linux-x86_64.zip
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0

100 1503k  100 1503k    0     0  [9](https://github.com/elastic/app-obs-dev/actions/runs/13620147754/job/38293121826?pr=153#step:3:11)912k      0 --:--:-- --:--:-- --:--:-- 9912k
Archive:  /tmp/tmp.IjQ6a7AYO1/artifact.zip
   creating: /tmp/tmp.IjQ6a7AYO1/out/jsonschema-6.0.3-linux-x86_64/bin/
  inflating: /tmp/tmp.IjQ6a7AYO1/out/jsonschema-6.0.3-linux-x86_64/bin/jsonschema  
install: cannot change permissions of ‘/usr/local/bin’: Operation not permitted
```

and with even previous versions

```
Run sourcemeta/jsonschema@ea5fc1a889f6c4d404e6c48109c37aaf9248b5b1
Run curl --retry [5](https://github.com/elastic/ingest-dev/actions/runs/13693574903/job/38291107179?pr=5196#step:3:7) --location --fail-early --silent --show-error \
  curl --retry 5 --location --fail-early --silent --show-error \
    --output "${GITHUB_WORKSPACE}/install.sh" \
    "https://raw.githubusercontent.com/sourcemeta/jsonschema/main/install"
  chmod +x "${GITHUB_WORKSPACE}/install.sh"
  "${GITHUB_WORKSPACE}/install.sh" 5.0.1
  rm "${GITHUB_WORKSPACE}/install.sh"
[...]

PR #250, opened 2025-03-06
fix: install in a different location for github actions

v1v:
Something has changed in the /usr/local/bin permission

However, GitHub actions can modify on the fly the PATH using the GITHUB_PATH env variable as explained in https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#adding-a-system-path

Part of https://github.com/sourcemeta/jsonschema/issues/249

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -6,9 +6,11 @@ runs:
     - name: Install the JSON Schema CLI (latest)
       shell: bash
       run: |
+        INSTALL_DIR="$HOME/.local/bin"
         curl --retry 5 --location --fail-early --silent --show-error \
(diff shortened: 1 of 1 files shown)


jviotti (approved):
Thank you so much!

kleinschmidt:
This doesn't seem to actually work in our workflows.  It's installing to `~/.local/bin/bin` but adding `~/.local/bin` to the path; fix in #251

jviotti:
Ah, good catch. Makes sense now that I see the scripts again. Please try v7.0.2. I just published it with the fix.

PR #251, opened 2025-03-06
Action fix: install dir should not contain `/bin`

kleinschmidt:
This is a followup to #250.  The install script seems to want the install root
and adds `/bin` to the install dir.  This means that the action installs the
binary to `${HOME}/.local/bin/bin` but adds `${HOME}/.local/bin` to the path,
which at least in my workflow is not enough to pick up the installed binary!

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -6,11 +6,11 @@ runs:
     - name: Install the JSON Schema CLI (latest)
       shell: bash
       run: |
-        INSTALL_DIR="$HOME/.local/bin"
+        INSTALL_DIR="$HOME/.local"
         curl --retry 5 --location --fail-early --silent --show-error \
           --output "${GITHUB_WORKSPACE}/install.sh" \
           "https://raw.githubusercontent.com/sourcemeta/jsonschema/main/install"
         chmod +x "${GITHUB_WORKSPACE}/install.sh"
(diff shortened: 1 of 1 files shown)


jviotti (approved):

jviotti:
Thanks! I'm on the go now, but I'll release a new version first thing tomorrow. You can still probably point the action at the specific commit SHA in `main` in the mean-time!

kleinschmidt:
Many thanks for the quick response and maintaining such a useful tool!!

jviotti:
Any time @kleinschmidt ! Out of curiosity, how are you using JSON Schema? If there is anything I can do on the CLI or other tooling to make your life easier, just let me know :)

kleinschmidt:
> Out of curiosity, how are you using JSON Schema?

We have some internal tools that our users need to write configuration files for, and validating that they conform to the schema is a great way to catch problems before we actually try to run a bad/incomplete/malformed config, and a great way to surface them in a human-readable way!  It also provides documentation for the config format that is co-located with the definition of the format and helps avoid our docs getting out of sync.

jviotti:
@kleinschmidt Very nice! As good error messages might be important there, JSON Schema is planning to work a bit on the whole idea of error reporting on GSoC 2025 (https://github.com/json-schema-org/community/issues/870). If you find this CLI (or any other validator) doesn't give you great error messages, please surface them as issues, as that will be a big area of focus soon.

PR #272, opened 2025-04-02
Support `pre-commit` with the `lint` command

kbroch-rivosinc:
I only tested the `jsonschema lint` command

Might be a lot more to consider here but at least this gets the conversation started.

For instance: 

* some projects choose to have separate repos to support pre-commit: https://github.com/astral-sh/ruff-pre-commit
* there are other ways to deploy hook besides `system` https://pre-commit.com/index.html#node

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,0 +1,31 @@
+- id: sourcemeta-jsonschema-fmt
+  name: sourcemeta-jsonschema-fmt
+  description: Format JSON or YAML schema files
+  entry: jsonschema fmt
+  language: system
+  types_or: [json, yaml]
+  minimum_pre_commit_version: 2.9.0
+
+- id: sourcemeta-jsonschema-lint
+  name: sourcemeta-jsonschema-lint
+  description: Lint JSON or YAML schema files
+  entry: jsonschema lint
+  language: system
+  types_or: [json, yaml]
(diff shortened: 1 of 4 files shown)


jviotti:
Thanks a lot! I'm off tomorrow but I'll check first thing on Friday.

.pre-commit-hooks.yaml
```diff
@@ -0,0 +1,31 @@
...
+  description: Validate that a JSON or YAML schema or a set of schemas are valid with their metaschema
+  entry: jsonschema metaschema
+  language: system
+  types_or: [json, yaml]
+  minimum_pre_commit_version: 2.9.0
+
+- id: sourcemeta-jsonschema-validate
+  name: sourcemeta-jsonschema-validate
+  description: Validate JSON or YAML files against a JSON or YAML schema
+  entry: jsonschema validate
```
jviotti: How do you pass options to this? i.e. which schema to validate against what?

jviotti:
> I only tested the jsonschema lint command

Also, if this is the only one you need, maybe let's just add that one first? We can merge, sort out the e2e testing, and then add others if/when needed?

jviotti:
Ah, I think I got it to work now. Let's see

jviotti:
Done! Thanks a lot! I'll merge and release this soon, and we can tweak the `pre-commit` definition to support any other command you need.

PR #283, opened 2025-04-07
Improve documentation for the `test` command

jviotti:

### DIFF
--- a/docs/test.markdown
+++ b/docs/test.markdown
@@ -20,10 +20,34 @@ suite](https://github.com/json-schema-org/JSON-Schema-Test-Suite).
 **If you want to validate that a schema adheres to its metaschema, use the
 [`metaschema`](./metaschema.markdown) command instead.**
 
+To test a schema, you define one or more test suite (i.e. collections of tests)
+as JSON files that follow a specific format:
+
+- `target`: The URI of the schema you want to test, which must be imported into
+  the resolution context using the `--resolve` or `--http` options
+- `tests`: An array of tests you want to run.
+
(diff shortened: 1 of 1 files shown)


docs/test.markdown
```diff
@@ -20,10 +20,34 @@ suite](https://github.com/json-schema-org/JSON-Schema-Test-Suite).
...
+
+- `target`: The URI of the schema you want to test, which must be imported into
+  the resolution context using the `--resolve` or `--http` options
+- `tests`: An array of tests you want to run.
+
+Every item in the `tests` array must be an object with the following
+properties:
+
+- `valid`: A boolean that determines whether the given instance is expected to
+  be valid or not against the target schema
```
adam-gol: "That is, if valid is FALSE, then if the data does not validate in the schema, the test passes.  If valid is TRUE, then if the data does validate in the schema, the test passes"

docs/test.markdown
```diff
@@ -20,10 +20,34 @@ suite](https://github.com/json-schema-org/JSON-Schema-Test-Suite).
...
+- `description`: An optional string property to make test output more readable
+
+And either of the following properties, but not both:
+
+- `data`: The instance that you want to test against the target schema
+- `dataPath`: The instance that you want to test against the target schema,
+  loaded from an external file instead
```
adam-gol, line 52: For example, if the schema under test has two fields required, "A" and "B", then example `data` would be:
`"data": {
   "fieldA": 123,
   "fieldB": "456"
}`

robbat2 (approved):

Issue #344, opened 2025-05-26
feature: Please ship CHECKSUMS, ideally GPG-signed


robbat2:
In the release process, please include CHECKSUMS of the distributed zipfiles.

Hash should be SHA256 or SHA512.

Optionally, but looks like it might be hard in your release process, the CHECKSUMS file should be GPG signed.

Example of JQ adding checksums - https://github.com/jqlang/jq/issues/726

jviotti:
Thanks for the suggestion. I'm up for doing it. Let's do it in two steps:

- I'll upload the checksums first, unsigned. Should be pretty easy
- I'm quite familiar with GPG. I can create an identity for i.e. Sourcemeta, trust the specific deploy release GitHub Actions job with the private key, and have the public key announced somewhere (maybe at sourcemeta.com) and linked to from this project's README

I'm travelling this week, so not sure how much of it will happen until next week, but let's go for it! 💪🏻 

Thanks for the suggestion as always

jviotti:
Here is an initial PR for the unsigned checksums: https://github.com/sourcemeta/jsonschema/pull/345

jviotti:
The unsigned checksums have been published: https://github.com/sourcemeta/jsonschema/releases/tag/v9.3.3. I'll aim to sort out the GPG signing next week.

jviotti:
It took some final GitHub Actions fiddling, but here we go! https://github.com/sourcemeta/jsonschema/releases/tag/v9.3.5. Thanks for the suggestion and let me know if you run into any problems with it!

PR #345, opened 2025-05-26
Publish (unsigned for now) SHA-512 checksums on GitHub Releases

jviotti:
See: https://github.com/sourcemeta/jsonschema/issues/344


.github/workflows/package.yml
```diff
@@ -120,6 +120,11 @@ jobs:
           path: build/out/
           pattern: artifacts-*
           merge-multiple: true
+
+      # Generate checksums for all pre-built binaries
+      - run: sha512sum build/out/*.zip > build/out/sha512sum.txt
```
robbat2: Suggest `--tag` for more useful output; and call the file `CHECKSUMS.txt` (sadly there is no standard for it)

.github/workflows/package.yml
```diff
@@ -130,15 +135,15 @@ jobs:
         run: |
           gh release delete continuous --cleanup-tag --yes || true
           gh release create continuous --prerelease --draft=false --title continuous --target "$GITHUB_SHA" --generate-notes
-          gh release upload --clobber continuous build/out/*.zip
+          gh release upload --clobber continuous build/out/*
```
robbat2: Is there a risk of other files in that directory? If yes, then explicitly keep the `*.zip` and add the specific checksum filename from the earlier step.

.github/workflows/package.yml
jviotti: Good idea!

.github/workflows/package.yml
jviotti: There shouldn't be any other files, but I guess it doesn't hurt to be explicit

PR #346, opened 2025-05-27
ci: Remove extra prefix on files in CHECKSUMS.txt

robbat2:
Fixes the path of files in the CHECKSUMS, so that the validation can be run directly.
```
SHA256(build/out/...) = ...
```

Reference: https://github.com/sourcemeta/jsonschema/issues/344
Signed-off-by: Robin H. Johnson <rjohnson@coreweave.com>

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -122,7 +122,7 @@ jobs:
           merge-multiple: true
 
       # Generate checksums for all pre-built binaries
-      # avoid "build/out/" in the checksum file.
+      # while avoiding "build/out/" in the checksum file.
       - run: sha256sum --tag *.zip >CHECKSUMS.txt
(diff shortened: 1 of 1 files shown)


.github/workflows/package.yml
```diff
@@ -122,8 +122,11 @@ jobs:
           merge-multiple: true
 
       # Generate checksums for all pre-built binaries
-      - run: sha256sum --tag build/out/*.zip > build/out/CHECKSUMS.txt
-      - run: cat build/out/CHECKSUMS.txt
+      # avoid "build/out/" in the checksum file.
```
jviotti:
```suggestion
      # while avoiding "build/out/" in the checksum file.
```

jviotti (approved):

PR #353, opened 2025-06-05
Sign `CHECKSUMS.txt` using a GPG key

jviotti:
See: https://github.com/sourcemeta/jsonschema/issues/344

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -121,12 +121,26 @@ jobs:
           pattern: artifacts-*
           merge-multiple: true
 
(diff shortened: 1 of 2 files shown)


robbat2:
Sorry; I didn't review in time. Would suggest `.asc` for the signature (--armor); as well as saying where to find the canonical key.

jviotti:
Good point on ASCII armour. Here you go: https://github.com/sourcemeta/jsonschema/pull/356. 

> as well as saying where to find the canonical key.

I had updated the README to mention the canonical location in the import command. Do you think that's enough?

PR #369, opened 2025-06-24
feat: publish the cli as snap

Karan-Palan:
For linux based operating systems, the cli must be published as a snap on the snap store for automatic updates
This PR adds Snap packaging and automatons:
1. snap/snapcraft.yaml
• Uses cmake plugin to build the C++ CLI on core22.
• Dynamically adopts the version from the most-recent Git tag so each new tag pushes a fresh Snap revision.
• Exposes a strict-confinement 
jsonschema
 app with home and network plugs.
2. .github/workflows/snap.yml
• Builds the snap on every version tag (v*.*.*) or manual dispatch.
• Stores the artifact and, when a tag build, publishes it to the Snap Store using a secret SNAPCRAFT_STORE_CREDENTIALS.


.github/workflows/snap.yml
```diff
@@ -0,0 +1,43 @@
...
+      # speed-up LXD initialisation which snapcraft uses under the hood
+      - name: Setup LXD
+        run: sudo snap install lxd && sudo lxd waitready && sudo lxd init --auto
+
+      # Build the snap in the default channel, outputs the path of the .snap file
+      - name: Build snap
+        id: snapcraft
+        uses: snapcore/action-build@v1
+        with:
+          path: ./snap
```
jviotti: How does it know how to build the project?

jviotti: Also, will this build the tool for both x64 and arm64? If not, is there a way to make it build for both?

.github/workflows/snap.yml
jviotti, line 1: Can we extend the `test` workflow with a new `snap` entry that builds the Snap (but doesn't publish it) so that we make sure to never break the integration?

jviotti: Also, can you add a new `snap` job on the `package.yml` action that will build the snap and build it as an artefact, and then extend the existing `publish` job to download the snap artefact and publish it (if it is a tag release)?

.github/workflows/snap.yml
Karan-Palan: > How does it know how to build the project?

snapcore/action-build@v1 is a gh action used to build snap packages. It will use instructions in the ./snap/snapcraft.yaml to build the snap

image

> Also, will this build the tool for both x64 and arm64? If not, is there a way to make it build for both?

Yes, now it does:
```
strategy:
      matrix:
        arch: [amd64, arm64]
```

.github/workflows/snap.yml
Karan-Palan: Done!

.github/workflows/snap.yml
jviotti: > It will use instructions in the ./snap/snapcraft.yaml to build the snap

I guess my question is how does it know how to build the project on the first place, as the YAML doesn't have specific CMake commands, etc. Assuming it's detecting the presence of `CMakeLists.txt` and just doing a standard CMake build, it would be good to have that explicitly set, so we can tweak the build params as needed?

.github/workflows/snap.yml
jviotti: This file can go away then?

snapcraft.yaml
```diff
@@ -0,0 +1,59 @@
...
+      - libstdc++-12-dev
+      - g++-12
+      - libc6-dev
+      - libc6-dev-i386
+      - g++-multilib
+    build-attributes: [keep-execstack, no-patchelf, no-install]
+
+    override-pull: |
+      craftctl default
+      VERSION=$(git describe --tags --abbrev=0 2>/dev/null || echo "v0.0.0")
```
jviotti: You should fetch the version from `CMakeLists.txt`.

snapcraft.yaml
```diff
@@ -0,0 +1,59 @@
...
+      craftctl set version="${VERSION#v}"
+
+    override-build: |
+      cmake -B build \
+        -DCMAKE_BUILD_TYPE=Release \
+        -DCMAKE_INSTALL_PREFIX=/usr \
+        -DBUILD_SHARED_LIBS=OFF \
+        -DCMAKE_EXE_LINKER_FLAGS="-static" \
```
jviotti: Why do we need these special flags? Was Snapcraft passing them by default?

snapcraft.yaml
```diff
@@ -0,0 +1,59 @@
...
+        -DCMAKE_EXE_LINK_DYNAMIC_C_FLAGS="-Wl,-Bstatic" \
+        -DCMAKE_EXE_LINK_DYNAMIC_CXX_FLAGS="-Wl,-Bstatic" \
+        -DCMAKE_SHARED_LIBRARY_LINK_C_FLAGS="-Wl,-Bstatic" \
+        -DCMAKE_SHARED_LIBRARY_LINK_CXX_FLAGS="-Wl,-Bstatic" \
+        $CRAFT_PART_SRC
+      cmake --build build --parallel "$(nproc)"
```
jviotti:
```suggestion
      cmake --build build --parallel "$(nproc)" --config Release
```

snapcraft.yaml
```diff
@@ -0,0 +1,59 @@
...
+      install -Dm755 build/src/jsonschema "$CRAFT_PART_INSTALL/bin/jsonschema"
```
jviotti: You should be using `cmake --install` rather than manually pulling the binary out with `install`

snapcraft.yaml
```diff
@@ -0,0 +1,59 @@
...
+
+    override-pull: |
+      craftctl default
+      VERSION=$(git describe --tags --abbrev=0 2>/dev/null || echo "v0.0.0")
+      craftctl set version="${VERSION#v}"
+
+    override-build: |
+      cmake -B build \
+        -DCMAKE_BUILD_TYPE=Release \
+        -DCMAKE_INSTALL_PREFIX=/usr \
```
jviotti: Should this be `$CRAFT_PART_INSTALL` given what you had below?

snapcraft.yaml
```diff
@@ -0,0 +1,59 @@
...
+    build-attributes: [keep-execstack, no-patchelf, no-install]
+
+    override-pull: |
+      craftctl default
+      VERSION=$(git describe --tags --abbrev=0 2>/dev/null || echo "v0.0.0")
+      craftctl set version="${VERSION#v}"
+
+    override-build: |
+      cmake -B build \
+        -DCMAKE_BUILD_TYPE=Release \
```
jviotti:
```suggestion
        -DCMAKE_BUILD_TYPE:STRING=Release \
```

snapcraft.yaml
jviotti: Actually this is invalid. You should only pass the prefix when calling `cmake --install` below

snapcraft.yaml
```diff
@@ -0,0 +1,59 @@
...
+        -DCMAKE_INSTALL_PREFIX=/usr \
+        -DBUILD_SHARED_LIBS=OFF \
```
jviotti:
```suggestion
        -DBUILD_SHARED_LIBS:BOOL=OFF \
```

snapcraft.yaml
```diff
@@ -0,0 +1,59 @@
+name: jsonschema
+base: core22
+adopt-info: jsonschema
+
+summary: Command-line tool for working with JSON Schema
+description: |
+  The JSON Schema CLI is a comprehensive command-line utility for validating,
```
jviotti: Can you use the exact same text as the GitHub repo description?

> The CLI for working with JSON Schema. Covers formatting, linting, testing, bundling, and more for both local development and CI/CD pipelines

.github/workflows/test.yml
```diff
@@ -109,3 +109,18 @@ jobs:
...
+      - name: Build snap
+        id: snapcraft
+        uses: snapcore/action-build@v1
+        with:
+          path: .
+      - name: Upload snap artefact
+        uses: actions/upload-artifact@v4
+        with:
+          name: snap-${{ github.sha }}.snap
+          path: ${{ steps.snapcraft.outputs.snap }}
```
jviotti: Can you also manually install the `.snap` file and run i.e. `jsonschema version` just to confirm it runs?

jviotti: Or even better, assuming `jsonschema` will end up in the path, try to run one or two representative test cases from `./test`, like:

```sh
./test/ci/pass_validate_http.sh jsonschema
```

.github/workflows/package.yml
```diff
@@ -110,20 +110,40 @@ jobs:
...
 
+  snap:
+    strategy:
+      matrix:
+        arch: [amd64, arm64]
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Setup LXD
+        run: sudo snap install lxd && sudo lxd waitready && sudo lxd init --auto
```
jviotti: You have this here but not on the `test` workflow. Can we remove it from here?

.gitignore
jviotti, line 1: The entire project build happens in `./build` to avoid polluting the project and having to ignore things out. Can you perform the build in `./build/snap`? Not sure if `snapcraft` lets you put an output directory, but otherwise you can `cd ./build/snap` and call snap craft from there

.github/workflows/test.yml
jviotti: Let me try this myself in the mean-time. I'll push a commit

.github/workflows/package.yml
```diff
@@ -110,20 +110,40 @@ jobs:
...
       - uses: actions/checkout@v4
       - uses: actions/download-artifact@v4
         with:
           path: build/out/
-          pattern: artifacts-*
+          pattern: "*"
           merge-multiple: true
 
-      # Generate signed checksums for all pre-built binaries
+      # Generate signed checksums for all pre-built binaries and snaps
```
jviotti: Snaps are prebuilt binaries anyway!

.github/workflows/package.yml
```diff
@@ -167,6 +187,17 @@ jobs:
         with:
           node-version: '22.x'
           registry-url: 'https://registry.npmjs.org'
+      - name: Publish Snap to Snap Store
```
jviotti: I propose we first publish the snaps to GitHub Releases. Once we make sure we got that right and that it all works well, we can look into publishing them to the official store. That way we can test all of this a few times before hitting the store for real

jviotti:
OK, I think I fixed everything. I'm trying the first GitHub Release deployment and we'll see how that goes

.github/workflows/snap.yml
Karan-Palan: It detected cMakeLists.txt and did a default build

> it would be good to have that explicitly set, so we can tweak the build params as needed?

Ahh, I see you have done that

snapcraft.yaml
Karan-Palan: Those were system specific, my local build was failing without them

Karan-Palan:
> See https://github.com/sourcemeta/jsonschema/releases/tag/v9.4.0 for an initial GitHub Release. Thanks a lot for this! Let's give it a shot this week and if it all works fine, we'll publish to the store

That's great news! I saw the changes you made, and everything's working perfectly now. It's much more simplified, which is a huge improvement. I wish I could've helped out, but I was asleep at the time. Anyway, I'm really looking forward to the CLI being available as a Snap and getting those automatic updates!

PR #371, opened 2025-06-26
Run an actual validating test for the new Ubuntu Snap

jviotti:
See: https://github.com/sourcemeta/jsonschema/pull/369


jviotti:
Judging by the error, I would say it cannot read files in directories other than `$HOME`, which I guess makes sense given the plugs. But we probably do need to allow it to read anywhere the user wants

Screenshot 2025-06-26 at 16 33 59

PR #409, opened 2025-07-21
chore: mention `mise` and `aqua` in the readme

mrazauskas:
Please, consider mentioning `mise` and `aqua` in the readme.

Reference: `jsonschema` was recently added to the `aqua` (https://github.com/aquaproj/aqua-registry/pull/39066) and `mise` (https://github.com/jdx/mise/pull/5714) registries.

---

Thanks for this project! I was looking for a tool like this one for some time.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -132,7 +132,7 @@ Where `X.Y.Z` is replaced with the desired version. For example:
 - run: jsonschema fmt path/to/schemas --check
 ```
 
-### From NPM
+### From npm
 
 ```sh
 npm install --global @sourcemeta/jsonschema
@@ -142,6 +142,12 @@ npm install --global @sourcemeta/jsonschema
 
 ```sh
 pip install sourcemeta-jsonschema
+```
+
+### From mise
+
+```sh
+mise use jsonschema
 ```
 
 ### From GitHub Releases


README.markdown
```diff
@@ -132,7 +132,7 @@ Where `X.Y.Z` is replaced with the desired version. For example:
 - run: jsonschema fmt path/to/schemas --check
 ```
 
-### From NPM
+### From npm
```
mrazauskas, line 135: This is how it is spelled allover the https://docs.npmjs.com. Also here: https://docs.npmjs.com/about-npm

README.markdown
```diff
@@ -144,6 +144,18 @@ npm install --global @sourcemeta/jsonschema
 pip install sourcemeta-jsonschema
 ```
 
+### From mise
+
+```sh
+mise use jsonschema
```
jviotti, line 150: Interesting. I never hears about these package managers. Are they specifically integrating with this tool? Might be worth putting a link to their homepages and the manifests that they have created for this tool, if any?

README.markdown
```diff
@@ -144,6 +144,18 @@ npm install --global @sourcemeta/jsonschema
...
+```
+
+### From aqua
+
+```sh
+aqua g -i sourcemeta/jsonschema
```
jviotti: Are you sure this works? I tried installing `aqua` from Homebrew and running this same command, but didn't work for me. Like above, maybe also worth linking to Aqua's homepage and the manifests that they have created for this tool, if any?

Finally, can we use the long form of the options instead of `g` and `-i`? That might better help readers understand

README.markdown
jviotti: Makes sense!

mrazauskas:
For the context, I use `mise` on daily basis. And `mise` is using registry from `aqua`. (It can also install from `cargo`, `pip`, `npm` or directly from GitHub releases and so on. These are known as different Backends.)

To make `mise use jsonschema` work, I had to include `jsonschema` to the `aqua` registry first. Although I do not use it.

The idea was to add just a short note that `jsonschema` is available from these tools as well. We can add links, of course. I was thinking, these notes only make sense for users of `mise` or `aqua`.

README.markdown
mrazauskas: Did you follow their [Quick Start](https://aquaproj.github.io/docs/tutorial) page? I also had troubles to setup `aqua`, because after installing one has to add it to the `PATH` manually and to run `aqua init`. And after `aqua g` one has to run `aqua i` again.

In their website they use short command names. I am not a fan of this too. Seems like `g` can be replaced with `generate`, but I can’t find a replacement for `-i`. (It means “insert“.)

Sounds complicated indeed. Perhaps this entry can be skipped?

jviotti: Yeah, makes sense. Maybe let's add the other one (and the NPM cleanup) and we try to research more about Aqua before adding it?

mrazauskas: Agreed.

jviotti:
Merged! Thanks a lot @mrazauskas. Out of curiosity, what are you using the tool for?

PR #413, opened 2025-07-25
Add new linting rules around `propertyNames`

jviotti:


jviotti:
cc @Karan-Palan Some of your new rules are landing here

PR #451, opened 2025-08-28
Add a `--loop` companion to the `--benchmark` option in the `validate` command

zx80:
This PR replaces #349.
It updates the code wrt the new way options are managed in the CLI.
I'm looking forward to see if the CI is happy with it…

### DIFF
--- a/docs/validate.markdown
+++ b/docs/validate.markdown
@@ -7,9 +7,9 @@ Validating
 ```sh
 jsonschema validate <schema.json|.yaml> <instance.json|.jsonl|.yaml...>
   [--http/-h] [--verbose/-v] [--resolve/-r <schemas-or-directories> ...]
-  [--benchmark/-b] [--extension/-e <extension>]
+  [--benchmark/-b] [--loop <iterations>] [--extension/-e <extension>]
   [--ignore/-i <schemas-or-directories>] [--trace/-t] [--fast/-f]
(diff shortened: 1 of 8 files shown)


jviotti:
I also added a few tests, and checked for `--loop` being greater than zero. The tests sometimes reveal the `nan` issue

jviotti:
Ah, interesting. OK. I added back the tests and just changed the C casts to `static_cast`. Let's see if CI passes!

jviotti:
Thanks a lot! I'll get this released in the next version!

jviotti:
See https://github.com/sourcemeta/jsonschema/releases/tag/v11.6.0. Thanks!

jviotti:
Looks like `--loop` is often flaky on CI. i.e. see https://github.com/sourcemeta/jsonschema/actions/runs/17552174327/job/49847076038. The problem is that the first time count may be negative, which I'm not sure it was intended:

```
/var/folders/y6/nj790rtn62lfktb1sh__79hc0000gn/T/tmp.ueHoR8ia2a/instance.json: PASS -0.018 +- 0.269 us (0.094)
```

I'll be taking a look, but if you have any ideas, let me know!

jviotti:
Maybe it makes sense to floor the duration on `0`? Just asking as I'm not sure how you are using these numbers

PR #454, opened 2025-08-29
jsonl: use --loop, show index and result

zx80:
add --loop handling with jsonl files as well.


src/command_validate.cc
```diff
@@ -70,6 +70,55 @@ auto parse_loop(const sourcemeta::core::Options &options) -> std::uint64_t {
...
+              const std::filesystem::path &instance_path,
+              const int64_t instance_index, const uint64_t loop) -> bool {
+  bool result = true;
+  double sum = 0.0, sum2 = 0.0, empty = 0.0;
+
+  // Overhead evaluation, if not to optimize out!
+  for (auto index = loop; index; index--) {
+    const auto start{std::chrono::high_resolution_clock::now()};
+    const auto end{std::chrono::high_resolution_clock::now()};
+    empty += (double)(std::chrono::duration_cast<std::chrono::nanoseconds>(
```
jviotti: Can you do `static_cast` as a more C++ idiom instead of `(double)` everywhere in the function?

src/command_validate.cc
```diff
@@ -141,7 +190,7 @@ auto sourcemeta::jsonschema::cli::validate(
       log_verbose(options)
           << "Interpreting input as JSONL: "
           << sourcemeta::core::weakly_canonical(instance_path).string() << "\n";
-      std::size_t index{0};
+      std::int64_t index{0};
```
jviotti, line 197: Is there a reason for this change? I guess `size_t` is more applicable, and it is an unsigned integer?

test/validate/pass_benchmark_loop_jsonl.sh
jviotti, line 1: Looks good,  but remember to register this file on `test/CMakeLists.txt`, otherwise it won't run!

src/command_validate.cc
zx80: Yes, this is to have -1 as a special value for simple json files vs jsonl files where a line is needed.

src/command_validate.cc
zx80: Dunno, I do not know about `static_cast`.
My knowledge of C++ is over 20 years old, and I was not planning to learn the language in depth right now. Having a few cast to avoid compiler warnings did not stuck me as very bad.

test/validate/pass_benchmark_loop_jsonl.sh
zx80: Ah. I'm as bad at cmake as at C++ :laughing:

zx80:
- I replace casts with `static_cast`, I'm not sure why the compiler would not figure that out on its own.
- I've added the test case to the list. Maybe this should be generated automatically instead of maintained manually.
- The CI complains about the code with windows-2022 native: After multiple failed blind attempts, I've no idea how to please the MSVC compiler, as I do not understand the error message.

jviotti:
Done! Thanks! I'll release a new version with both PRs in a bit

PR #471, opened 2025-09-06
feat(version): support -v command

### LINKED ISSUE #470: Support `-v` to show the cli version
As of version 11.8.0, we can only check the cli version using the `--version` command, It'd be great to support a shorter `-v` command to show the version, as its consistent with other packages and cli's:

Image

Karan-Palan:
Fixes #470 

Before and after:

image

Added tests, updated relevant docs

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -15,7 +15,7 @@
 constexpr std::string_view USAGE_DETAILS{R"EOF(
 Global Options:
 
-   --verbose, -v                  Enable verbose output
+   --verbose                      Enable verbose output
    --resolve, -r                  Import the given JSON Schema (or directory of schemas)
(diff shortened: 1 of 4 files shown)


src/main.cc
```diff
@@ -15,15 +15,15 @@
 constexpr std::string_view USAGE_DETAILS{R"EOF(
 Global Options:
 
-   --verbose, -v                  Enable verbose output
```
jviotti: I think we shouldn't remove `-v` here, as it is still applicable as a verbosity flag in most other commands.

jviotti:
The PR looks good. The failures seem unrelated. CI tests that hit an HTTP server. Let me fix those in a separate PR

PR #524, opened 2025-10-29
Support printing errors as JSON given `--json`

jviotti:


jviotti:
@Karan-Palan Its happening!

Issue #304, opened 2025-04-17
Core dump?


adam-gol:
Brand new install,
```
root ➜ /workspaces/x590schema (main) $ uname -a
Linux e868adb3c5a2 5.15.167.4-microsoft-standard-WSL2 #1 SMP Tue Nov 5 00:21:55 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux
```

Seems to be a good binary,
```
root ➜ /workspaces/x590schema (main) $ jsonschema
JSON Schema CLI - v9.1.0
Usage: jsonschema <command> [arguments...]

Global Options:

   --verbose, -v                  Enable verbose output
   --resolve, -r                  Import the given JSON Schema (or directory of schemas)
                                  into the resolution context
   --default-dialect, -d <uri>    Specify the URI for the default dialect to be used
                                  if the `$schema` keyword is not set

```

and yet, 

```
root ➜ /workspaces/x590schema (main) $ jsonschema test test/test.json --resolve  -v
Illegal instruction (core dumped)
.json --resolve x590schema.json -vn) $ jsonschema validate example-subschema.json example-instance.json --resolve x590signatures.
Illegal instruction (core dumped)
```

Maybe I'm doing something stupid.

jviotti:
Hey @adam-gol , that's very weird. We run tests on CI on Linux of course. 

- Is this the same machine where you previously ran the CLI just fine?
- Is this happening on a specific version or all of them? (https://github.com/sourcemeta/jsonschema/releases/tag/v9.1.0)
- Is this on the WSL on Windows? Which distro? Is there a way I can try exactly the same setup?
- Only if you are somewhat familiar with build systems, can you try building the CLI from source? Assuming you have `cmake`, you should be able to just run `make` and run the binary from `.bin/jsonschema`. I wonder if any of our build settings made the CLI incompatible with your Linux runtime?

adam-gol:
Shrug.

The machine this is running on is Win7, Docker 4.33.1 (161083), 
mcr.microsoft.com/devcontainers/base:jammy 

Happens in 9.0.0 and 9.1.0 (at least).

adam-gol:
```
[  917.056749] WSL (8623 - CaptureCrash): Capturing crash for pid: 2922, executable: !usr!bin!jsonschema, signal: 4, port: 50005
[...]

PR #526, opened 2025-10-30
Dockerfile: also disable SIMD support on Docker binaries as well

peat-psuwit:
Docker binaries are being used in a similar way as CI binaries, so they should receive the same treatment.

While we're at it, add line breaks.

See: https://github.com/sourcemeta/jsonschema/issues/304


jviotti (approved):
Ahh great catch. Thank you very much! I will likely release a new version tomorrow

PR #551, opened 2025-11-21
Do not stop after the first failure in `fmt` and support `--json`

jviotti:


Karan-Palan:
Yes, also better for the diff!

PR #564, opened 2025-12-02
Expose a spawning interface in the CLI NPM wrapper

jviotti:


jviotti:
cc @Karan-Palan For the extension

PR #570, opened 2025-12-03
Support cases where platform.machine() is 'amd64' instead of 'x86_64'

EMBailey-Topaz:
This updates `__main__.py` in the PyPi version to work when `platform.system().lower()` returns `'amd64'`.


jviotti (approved):
Awesome. Thanks!

PR #599, opened 2026-01-05
feat: order lint errors by line number

Karan-Palan:
Fixes https://github.com/sourcemeta/studio/issues/140 based on https://github.com/sourcemeta/studio/pull/141#issuecomment-3702166632

Updated existing tests that were failing after the changes to demonstrate working

### DIFF
--- a/src/command_lint.cc
+++ b/src/command_lint.cc
@@ -299,22 +299,8 @@ auto sourcemeta::jsonschema::lint(const sourcemeta::core::Options &options)
     std::sort(errors_array.as_array().begin(), errors_array.as_array().end(),
               [](const sourcemeta::core::JSON &left,
                  const sourcemeta::core::JSON &right) {
-                const auto left_path_str = left.at("path").to_string();
-                const auto right_path_str = right.at("path").to_string();
-                if (left_path_str != right_path_str) {
-                  return left_path_str < right_path_str;
-                }
-
-                const auto &left_pos = left.at("position");
-                const auto &right_pos = right.at("position");
-
-                if (left_pos.is_null())
-                  return false;
-                if (right_pos.is_null())
-                  return true;
-
-                return left_pos.front().to_integer() <
-                       right_pos.front().to_integer();
+                return left.at("position").front() <
+                       right.at("position").front();
               });
 
     auto output_json_object = sourcemeta::core::JSON::make_object();


src/command_lint.cc
```diff
@@ -296,6 +296,54 @@ auto sourcemeta::jsonschema::lint(const sourcemeta::core::Options &options)
   }
 
   if (output_json) {
+    std::vector<std::size_t> indices(errors_array.size());
+    for (std::size_t i = 0; i < errors_array.size(); ++i) {
+      indices[i] = i;
+    }
+
+    std::sort(indices.begin(), indices.end(),
```
jviotti: This looks very complicated. You should be able to sort within the JSON Array itself:

```
std::ranges::sort(errors_array.as_array(), ....);
```

And for the ordering criteria, what about just `a.at("position").front() > b.at("position").front()` or something like that?

src/command_lint.cc
```diff
@@ -296,6 +296,27 @@ auto sourcemeta::jsonschema::lint(const sourcemeta::core::Options &options)
   }
 
   if (output_json) {
+    std::sort(errors_array.as_array().begin(), errors_array.as_array().end(),
+              [](const sourcemeta::core::JSON &left,
+                 const sourcemeta::core::JSON &right) {
+                const auto left_path_str = left.at("path").to_string();
```
jviotti: Do we really need to sort on paths? I think just sorting on line numbers will work?

src/command_lint.cc
```diff
@@ -296,6 +296,27 @@ auto sourcemeta::jsonschema::lint(const sourcemeta::core::Options &options)
...
+                const auto right_path_str = right.at("path").to_string();
+                if (left_path_str != right_path_str) {
+                  return left_path_str < right_path_str;
+                }
+
+                const auto &left_pos = left.at("position");
+                const auto &right_pos = right.at("position");
+
+                if (left_pos.is_null())
```
jviotti: I think you might not need to handle this case. The JSON class should support comparison even on different types

src/command_lint.cc
```diff
@@ -296,6 +296,27 @@ auto sourcemeta::jsonschema::lint(const sourcemeta::core::Options &options)
...
+                  return false;
+                if (right_pos.is_null())
+                  return true;
+
+                return left_pos.front().to_integer() <
```
jviotti: Same here. I don't think you need the `to_integer()`. The JSON class should be able to handle it

jviotti:
Getting released here: https://github.com/sourcemeta/jsonschema/actions/runs/20749102883

PR #605, opened 2026-01-10
feat: URL resolving support for validate & metaschema

### LINKED ISSUE #234: Extend `validate`/`metaschema` to directly take a URL to a schema to validate

robbat2:
Closes: https://github.com/sourcemeta/jsonschema/issues/234
Co-authored-by: Codex <codex@openai.com>
Generated-with: OpenAI Codex CLI (partial)
Signed-off-by: Robin H. Johnson <rjohnson@coreweave.com>


jviotti (changes requested):
I love the idea. I think the way Codex/you are implementing it is fighting the current machinery for iterating on schema arguments. I think we need for make `for_each_json` transparently work with both URLs or file paths. The `src/input.h` file abstracts a way a lot of the logic, including line number gathering, reading, etc so that the actual commands never have to worry about it.

`for_each_json` will loop and pass specific positional arguments to `handle_json_entry`. It feels to me that such `handle_json_entry` needs to take either a path or a URI (maybe just a string?) and then internally check: if it is a path and it exists and it is a file, do X, if it is a directory, do Y, if it is an HTTP/S URL, do Z.

I think if we do all of that there, we very elegantly gain URL support in every applicable command.

jviotti:
If you merge main back in, the transient CI issue should be gone!

src/input.h
```diff
@@ -26,14 +29,40 @@
...
+
+  [[nodiscard]] auto is_local() const noexcept -> bool {
+    return this->path.has_value();
+  }
+
+  [[nodiscard]] auto local_path_or_throw(std::string_view command) const
+      -> const std::filesystem::path & {
+    if (!this->path.has_value()) {
+      std::ostringstream error;
+      error << "The `" << command << "` command does not support HTTP inputs";
```
jviotti, line 52: Note we never dynamically construct strings for exceptions. Instead, use one of the existing error classes that would be handled in `src/error.h` or create a new exception class there

src/input.h
```diff
@@ -26,14 +29,40 @@
...
+      throw std::runtime_error(error.str());
+    }
+
+    return this->path.value();
+  }
+
+  [[nodiscard]] auto base_uri() const -> std::string {
+    if (sourcemeta::jsonschema::is_http_url(this->first)) {
```
jviotti, line 60:
```suggestion
    if (is_http_url(this->first)) {
```

You are already in the `sourcemeta::jsonschema::` namespace, so you can omit it in every case.

src/input.h
```diff
@@ -26,14 +29,40 @@
...
+    if (!this->path.has_value()) {
+      std::ostringstream error;
+      error << "The `" << command << "` command does not support HTTP inputs";
+      throw std::runtime_error(error.str());
+    }
+
+    return this->path.value();
+  }
+
+  [[nodiscard]] auto base_uri() const -> std::string {
```
jviotti, line 59: This method is called `base_uri` but it doesn't return a _base_ URI, but the whole file URI given the path?

src/input.h
```diff
@@ -26,14 +29,40 @@
 namespace sourcemeta::jsonschema {
 
 struct InputJSON {
-  std::filesystem::path first;
+  std::string first;
```
jviotti, line 32: If the thing here is that the `first` might be _either_ a path or a URI, maybe a `std::variant` expresses this better

src/input.h
```diff
@@ -144,16 +173,132 @@ inline auto read_file(const std::filesystem::path &path) -> ParsedJSON {
...
+          "Remote schema inputs require network access. Pass `--http/-h`"};
+    }
+
+    const cpr::Response response{
+        sourcemeta::jsonschema::fetch_http_response(options, entry_identifier)};
+
+    const auto path_without_query_or_fragment =
+        sourcemeta::jsonschema::uri_path_without_query_or_fragment(
+            entry_identifier);
+    if (path_without_query_or_fragment.ends_with(".jsonl")) {
```
jviotti, line 193: Beyond just the extension, I think we need to look at the content type mime header?

src/http.h
```diff
@@ -0,0 +1,133 @@
...
+inline auto is_http_url(std::string_view identifier) -> bool {
+  const sourcemeta::core::URI uri{std::string{identifier}};
+  const auto maybe_scheme{uri.scheme()};
+  return maybe_scheme.has_value() &&
+         (maybe_scheme.value() == "https" || maybe_scheme.value() == "http");
+}
+
+inline auto uri_path_without_query_or_fragment(std::string_view uri)
+    -> std::string_view {
+  const auto fragment = uri.find('#');
```
jviotti, line 41: You can use the `sourcemeta::core::URI` class instead of manual string manipulation. That class handles all the edge cases properly. There is a `recompose_without_fragment` method for this

src/http.h
```diff
@@ -0,0 +1,133 @@
...
+    }
+  }
+
+  const auto path = uri_path_without_query_or_fragment(uri);
+  return path.ends_with(".yaml") || path.ends_with(".yml");
+}
+
+inline auto fetch_http_response(const sourcemeta::core::Options &options,
+                                std::string_view uri) -> cpr::Response {
+  if (!is_http_url(uri)) {
```
jviotti, line 73: I think you guard against that when calling the function, so maybe unnecessary to check again here? Could be just an `assert`

src/http.h
```diff
@@ -0,0 +1,133 @@
...
+    throw std::runtime_error{"The input is not an HTTP URL"};
+  }
+
+  LOG_VERBOSE(options) << "Fetching over HTTP: " << uri << "\n";
+  const cpr::Response response{
+      cpr::Get(cpr::Url{std::string{uri}}, cpr::Redirect{true})};
+
+  if (response.status_code < 200 || response.status_code >= 300) {
```
jviotti, line 81: We had implemented a retry logic on `main`, which I think got lost in this branch?

src/http.h
```diff
@@ -0,0 +1,133 @@
...
+}
+
+inline auto fetch_http_schema(const sourcemeta::core::Options &options,
+                              std::string_view uri) -> sourcemeta::core::JSON {
+  auto official_result{sourcemeta::core::schema_resolver(uri)};
+  if (official_result.has_value()) {
+    return official_result.value();
+  }
+
+  if (!is_http_url(uri)) {
```
jviotti, line 124: Can this be an assert?

src/http.h
```diff
@@ -0,0 +1,133 @@
...
+            std::move(positions), false};
+  } catch (const sourcemeta::core::JSONParseError &error) {
+    throw RemoteSchemaJSONParseError{std::string{uri}, error.line(),
+                                     error.column(), error.what()};
+  }
+}
+
+inline auto fetch_http_schema(const sourcemeta::core::Options &options,
+                              std::string_view uri) -> sourcemeta::core::JSON {
+  auto official_result{sourcemeta::core::schema_resolver(uri)};
```
jviotti, line 119: This looks a bit weird here inside a `fetch_http_schema` function. I think callers should be doing this instead and if no result, _then_ call `fetch_http_schema`?

src/error.h
```diff
@@ -142,6 +142,75 @@ class Fail : public std::runtime_error {
   int exit_code_;
 };
 
+class RemoteSchemaFetchError : public std::runtime_error {
+public:
+  RemoteSchemaFetchError(std::string uri, const long status_code)
+      : std::runtime_error{"Could not fetch the schema over HTTP (HTTP " +
```
jviotti, line 148: Let's not dynamically construct errors please. The idea is that the dynamic components, like in this case the status code, should be a method of the exception, so we report it as a separate property when converting the exception to JSON when `--json` is passed

src/error.h
```diff
@@ -142,6 +142,75 @@ class Fail : public std::runtime_error {
...
+
+  [[nodiscard]] auto uri() const noexcept -> const std::string & {
+    return this->uri_;
+  }
+
+private:
+  std::string uri_;
+};
+
+class RemoteSchemaJSONParseError : public std::runtime_error {
```
jviotti, line 160: A lot of these errors already exist. I think we need a `RemoteError` equivalent to `FileError`? We added the latter so we could wrap any existing exception with file information instead of re-creating many errors over and over again (like this)

test/ci/fail_metaschema_schema_http_404.sh
```diff
@@ -0,0 +1,49 @@
...
+  sleep 0.1
+done
+PORT="$(cat "$PORT_FILE")"
+
+"$1" metaschema "http://127.0.0.1:$PORT/missing.json" --http \
+  2>"$TMP/stderr.txt" && CODE="$?" || CODE="$?"
+test "$CODE" = "1" || exit 1
+
+cat << EOF > "$TMP/expected.txt"
+error: Could not fetch the schema over HTTP (HTTP 404)
```
jviotti, line 45: For every test that showcases an error, can you also add after the text assert a `--json` assert? See the other tests. We usually do this in all `fail_` files to make sure both JSON and non-JSON variants of the error are properly reported

jviotti: And we probably want the status to be a member itself, like `with status 404`?

test/ci/pass_metaschema_schema_http.sh
```diff
@@ -0,0 +1,66 @@
...
+  TRIES=$((TRIES + 1))
+  if [ "$TRIES" -gt 100 ]; then
+    cat "$SERVER_LOG" >&2 || true
+    exit 1
+  fi
+  sleep 0.1
+done
+PORT="$(cat "$PORT_FILE")"
+
+"$1" metaschema "http://127.0.0.1:$PORT/schema.json" --http \
```
jviotti, line 60: Maybe run this one with `--verbose` here? So we capture that output to make sure it did indeed work?

test/ci/pass_validate_schema_http.sh
```diff
@@ -0,0 +1,70 @@
...
+  fi
+  sleep 0.1
+done
+PORT="$(cat "$PORT_FILE")"
+
+cat << 'EOF' > "$TMP/instance.json"
+{ "name": "foo", "kind": "example" }
+EOF
+
+"$1" validate "http://127.0.0.1:$PORT/schema.json" "$TMP/instance.json" --http \
```
jviotti, line 64: I'm now wondering if a command invocation to an HTTP URL should assume `--http`. I think it makes sense?

test/ci/pass_validate_schema_http_suffix_json_over_plain_content_type.sh
```diff
@@ -0,0 +1,63 @@
...
+PORT="$(cat "$PORT_FILE")"
+
+cat << 'EOF' > "$TMP/instance.json"
+{ "name": "foo" }
+EOF
+
+"$1" validate "http://127.0.0.1:$PORT/schema.json?ignored=1" "$TMP/instance.json" \
+  --http 2>"$TMP/stderr.txt"
+
+cat << EOF > "$TMP/expected.txt"
```
jviotti, line 60: Can we run all of these that present no output with `--verbose` and actually capture BOTH stdout and stderr? i.e. if for whatever reason this command printed non-sense to `stdout`, we would never know

test/lint/pass_lint_file_uri.sh
```diff
@@ -0,0 +1,33 @@
...
+  "title": "Test",
+  "description": "Test schema",
+  "enum": [ "foo" ]
+}
+EOF
+
+SCHEMA_PATH="$(cd "$TMP" && pwd)/schema.json"
+SCHEMA_URI="file://$SCHEMA_PATH"
+
+"$1" lint "$SCHEMA_URI" --json >"$TMP/output.json" 2>&1
```
jviotti, line 22: Maybe add a `_json` suffix to the test file as this is only covering the JSON part?

test/metaschema/fail_schema_url_without_http.sh
jviotti, line 1: I get this test, but I'm wondering now if it is a bit confusing. I think if any input is an HTTP URL, we should assume `--http`, as that's clearly the user's intent?

test/metaschema/pass_file_uri.sh
```diff
@@ -0,0 +1,28 @@
...
+SCHEMA_PATH="$(cd "$TMP" && pwd)/schema.json"
+SCHEMA_URI="file://$SCHEMA_PATH"
+
+"$1" metaschema "$SCHEMA_URI" > "$TMP/output.txt" 2>&1
+
+cat << EOF > "$TMP/expected.txt"
+EOF
+
+diff "$TMP/output.txt" "$TMP/expected.txt"
+
```
jviotti, line 28: Looks like various test files have this empty line here?

jviotti (changes requested):
I left some comments. I also wonder if there is a way to split up this PR so it's easy to merge the core logic first and avoid git conflicts, but not sure what that would be 🤔

robbat2:
@jviotti your codebase has moved on since this was opened; I'm going to close this as-is and let you take what's useful for later. Would still like URL resolving, but it's less important than I originally thought.

Issue #537, opened 2025-11-13
Allow all commands to take input using standard input


jviotti:
For https://github.com/sourcemeta/studio/issues/61

PR #614, opened 2026-01-15
CLI: add stdin support for instance validation 

Vaibhav701161:
Adds stdin support for instance input to enable real-time editor diagnostics.
Fixes(#537)
* Centralized stdin parsing in `src/input.h` (JSON-first, YAML fallback).
* `validate` accepts instances from stdin only; schemas must be file paths.
* Stdin inputs resolve relative `$ref` against the current working directory.
* Includes integration test `test/validate/pass_stdin.sh`.
* Unblocks `sourcemeta/studio#61`; addresses `sourcemeta/jsonschema#537`.

Refs: sourcemeta/studio#61, sourcemeta/jsonschema#537

*`Build and test:`
Screenshot 2026-01-15 145159


src/input.h
```diff
@@ -143,12 +144,42 @@ inline auto read_file(const std::filesystem::path &path) -> ParsedJSON {
   }
 }
 
+inline auto read_from_stdin() -> ParsedJSON {
+  std::string buffer;
+  {
+    std::ostringstream ss;
```
jviotti: You shouldn't need this. `sourcemeta::core::parse_json` should already be able to read from any standard C++ input stream without needing to buffer it

src/input.h
```diff
@@ -27,6 +27,7 @@ namespace sourcemeta::jsonschema {
 
 struct InputJSON {
   std::filesystem::path first;
+  std::filesystem::path resolution_base;
```
jviotti, line 31: If this is just a display name, maybe make a `std::string`? Otherwise `<stdin>` is not really a path

test/validate/pass_stdin.sh
jviotti, line 1: Please split this test into multiple tests. If you see the other tests in the project, we have specific conventions and we always test a single thing per file. Plus you also want to add test for every other applicable command.

Finally, you didn't register this test in `test/CMakeLists.txt`, so it won't run at all

src/command_lint.cc
```diff
@@ -210,7 +210,8 @@ auto sourcemeta::jsonschema::lint(const sourcemeta::core::Options &options)
               const auto apply_result = bundle.apply(
                   copy, sourcemeta::core::schema_walker, custom_resolver,
                   get_lint_callback(errors_array, entry, output_json), dialect,
-                  sourcemeta::jsonschema::default_id(entry.first),
+                  sourcemeta::core::URI::from_path(entry.resolution_base)
```
jviotti: I think we still definitely want to use `sourcemeta::jsonschema::default_id` here, to concentrate that logic in a single place. Maybe make such function take `entry` directly, so that inside we can decide to use `entry.resolution_base` or whatever?

src/command_lint.cc
```diff
@@ -282,7 +283,8 @@ auto sourcemeta::jsonschema::lint(const sourcemeta::core::Options &options)
                   entry.second, sourcemeta::core::schema_walker,
                   custom_resolver,
                   get_lint_callback(errors_array, entry, output_json), dialect,
-                  sourcemeta::jsonschema::default_id(entry.first),
+                  sourcemeta::core::URI::from_path(entry.resolution_base)
```
jviotti: Same here

src/input.h
jviotti: I think this still needs to be addressed?

src/input.h
jviotti: This too is still valid

src/input.h
```diff
@@ -143,12 +144,42 @@ inline auto read_file(const std::filesystem::path &path) -> ParsedJSON {
...
+
 inline auto
 handle_json_entry(const std::filesystem::path &entry_path,
                   const std::set<std::filesystem::path> &blacklist,
                   const std::set<std::string> &extensions,
                   std::vector<sourcemeta::jsonschema::InputJSON> &result,
                   const sourcemeta::core::Options &options) -> void {
+  if (entry_path == "-") {
+    auto parsed{read_from_stdin()};
+    std::filesystem::path display_path{"<stdin>"};
```
jviotti: I don't think we need to declare these as separate variables and then move them. Probably worth doing `.emplace_back` with all the data directly?

src/input.h
```diff
@@ -171,7 +202,9 @@ handle_json_entry(const std::filesystem::path &entry_path,
 
         // TODO: Print a verbose message for what is getting parsed
         auto parsed{read_file(canonical)};
-        result.push_back({std::move(canonical), std::move(parsed.document),
+        auto canonical_copy{canonical};
```
jviotti: Do we really need this copy as a variable? Can you just pass them both directly to the `push_back`?

src/input.h
```diff
@@ -240,33 +273,46 @@ handle_json_entry(const std::filesystem::path &entry_path,
...
-            result.push_back({canonical, std::move(entry.first),
+            result.push_back({canonical, canonical, std::move(entry.first),
                               std::move(entry.second), index, true, true});
             index += 1;
           }
         } else if (documents.size() == 1) {
-          result.push_back(
-              {std::move(canonical), std::move(documents.front().first),
-               std::move(documents.front().second), 0, false, true});
+          auto canonical_copy{canonical};
```
jviotti: Same here. Can we avoid the intermediate variable?

src/input.h
```diff
@@ -240,33 +273,46 @@ handle_json_entry(const std::filesystem::path &entry_path,
...
+                            true});
         }
       } else {
         if (std::filesystem::is_empty(canonical)) {
           return;
         }
         // TODO: Print a verbose message for what is getting parsed
         auto parsed{read_file(canonical)};
-        result.push_back({std::move(canonical), std::move(parsed.document),
+        auto canonical_copy{canonical};
```
jviotti: Same here

src/input.h
```diff
@@ -240,33 +273,46 @@ handle_json_entry(const std::filesystem::path &entry_path,
...
+        result.push_back({std::move(canonical), std::move(canonical_copy),
+                          std::move(parsed.document),
                           std::move(parsed.positions), 0, false, parsed.yaml});
       }
     }
   }
 }
-
```
jviotti, line 264: I don't think this deletion was intended?

src/input.h
```diff
@@ -240,33 +273,46 @@ handle_json_entry(const std::filesystem::path &entry_path,
...
                           const sourcemeta::core::Options &options)
     -> std::vector<InputJSON> {
+  std::size_t stdin_count{0};
+  for (const auto &arg : arguments) {
+    if (arg == "-") {
+      stdin_count++;
+    }
+  }
+  if (stdin_count > 1) {
+    throw std::runtime_error{"Standard input (-) can only be specified once"};
```
jviotti:
```suggestion
    throw std::runtime_error{"The standard input position argument `-` can only be specified once"};
```

src/input.h
```diff
@@ -240,33 +273,46 @@ handle_json_entry(const std::filesystem::path &entry_path,
...
-
 } // namespace
 
 inline auto for_each_json(const std::vector<std::string_view> &arguments,
                           const sourcemeta::core::Options &options)
     -> std::vector<InputJSON> {
+  std::size_t stdin_count{0};
+  for (const auto &arg : arguments) {
+    if (arg == "-") {
+      stdin_count++;
```
jviotti: I don't think counting all occurrences makes sense if you only want up to two. Maybe check that if you already got at least 2, then its an error and you can break?

test/validate/pass_stdin.sh
```diff
@@ -0,0 +1,59 @@
...
+EOF
+
+echo '{"foo": "bar"}' | "$1" validate "$TMP/schema.json" -
+
+echo 'foo: bar' | "$1" validate "$TMP/schema.json" -
+
+set +o errexit
+cat "$TMP/schema.json" > "$TMP/schema_stdin.json"
+
+OUTPUT=$("$1" validate - "$TMP/schema.json" < "$TMP/schema_stdin.json" 2>&1)
```
jviotti: Please see the other shell scripts for the conventions on how we test stuff. Some obvious ones:

- We never capture into a variable. We pipe into a file and then `diff` it
- We never enable `set +o errexit`
- We never test more than one thing at once

Also your test is not registered in `test/CMakeLists.txt` so it won't run at all. But overall, split it so that there are multiple tests testing each one thing

src/input.h
Vaibhav701161: I removed it and now pass std::cin directly to parse_json.

src/command_fmt.cc
```diff
@@ -21,21 +21,26 @@ auto sourcemeta::jsonschema::fmt(const sourcemeta::core::Options &options)
   std::vector<std::string> failed_files;
   const auto indentation{parse_indentation(options)};
   for (const auto &entry : for_each_json(options)) {
+    if (entry.from_stdin) {
+      throw std::runtime_error{
```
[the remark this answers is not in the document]
Vaibhav701161: No existing StdinNotSupportedError or equivalent in error.h . Introducing a new one would exceed scope.

src/input.h
```diff
@@ -143,12 +146,26 @@ inline auto read_file(const std::filesystem::path &path) -> ParsedJSON {
   }
 }
 
+inline auto read_from_stdin() -> ParsedJSON {
+  sourcemeta::core::PointerPositionTracker positions;
+  return {sourcemeta::core::parse_json(std::cin, std::ref(positions)),
```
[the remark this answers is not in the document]
Vaibhav701161: Would require buffering stdin

test/validate/fail_stdin_multiple.sh
```diff
@@ -0,0 +1,25 @@
...
+
+echo '{}' | "$1" validate "$TMP/schema.json" - - 2>"$TMP/stderr.txt" \
+  && EXIT_CODE="$?" || EXIT_CODE="$?"
+test "$EXIT_CODE" = "1" || exit 1
+
+cat << 'EOF' > "$TMP/expected.txt"
+error: The standard input position argument `-` can only be specified once
+EOF
+
+diff "$TMP/stderr.txt" "$TMP/expected.txt"
```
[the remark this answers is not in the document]
Vaibhav701161: I will fix this.

test/validate/fail_stdin_schema.sh
```diff
@@ -0,0 +1,22 @@
...
+
+"$1" validate - "$TMP/instance.json" 2>"$TMP/stderr.txt" \
+  && EXIT_CODE="$?" || EXIT_CODE="$?"
+test "$EXIT_CODE" = "1" || exit 1
+
+cat << 'EOF' > "$TMP/expected.txt"
+error: Reading the schema from standard input is not supported
+EOF
+
+diff "$TMP/stderr.txt" "$TMP/expected.txt"
```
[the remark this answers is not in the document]
Vaibhav701161: I will fix this.

test/validate/fail_stdin_multiple.sh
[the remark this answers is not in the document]
jviotti: @Vaibhav701161 This is a valid one. Needs fixing!

test/validate/fail_stdin_schema.sh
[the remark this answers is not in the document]
jviotti: Same here.

test/validate/pass_stdin_instance.sh
jviotti, line 1: Can you add a failing case too?

src/command_validate.cc
```diff
@@ -141,6 +141,11 @@ auto sourcemeta::jsonschema::validate(const sourcemeta::core::Options &options)
 
   const auto &schema_path{options.positional().at(0)};
 
+  if (schema_path == "-") {
+    throw std::runtime_error{
+        "Reading the schema from standard input is not supported"};
```
jviotti, line 146: Why not? Sounds like a valid use case, unless there is another reason why it wouldn't work that I'm not thinking about?

test/CMakeLists.txt
jviotti, line 1: So far we only have `stdin` tests for the `validate` command. But we should be testing stdin behaviour for every command that supports it. Otherwise it could be totally broken in other commands and we are not knowing about it

src/command_fmt.cc
```diff
@@ -21,21 +21,26 @@ auto sourcemeta::jsonschema::fmt(const sourcemeta::core::Options &options)
...
+          "This command does not support reading from standard input"};
```
jviotti, line 26: Why not?

PR #628, opened 2026-01-26
Add VS code extension promotion message on install across various distributions

Karan-Palan:
#### npm
- Added `postinstall` script to `package.json`
- Created `npm/postinstall.js` that displays extension message after installation

#### CLI Binary
- Modified `src/main.cc` help command to show extension link
- Appears when running `jsonschema`, `jsonschema --help`, or `jsonschema -h`

#### Install Script (Homebrew/Direct install)
- Updated `install` script to show message after successful installation

Screenshot from 2026-01-25 19-00-41

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -210,13 +210,6 @@ auto jsonschema_main(const std::string &program, const std::string &command,
     std::cout << "Usage: " << std::filesystem::path{program}.filename().string()
               << " <command> [arguments...]\n";
     std::cout << USAGE_DETAILS;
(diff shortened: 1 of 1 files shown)


src/main.cc
jviotti, line 1: I like the other ones but listing it every time on the help page might be too much. Instead of in the help, maybe let's also put it on the Homebrew cask? https://github.com/sourcemeta/homebrew-apps/blob/main/Casks/jsonschema.rb

Karan-Palan: Raised a PR - https://github.com/sourcemeta/homebrew-apps/pull/3

PR #648, opened 2026-02-05
Better distinguish error categories using different exit codes

RitoG09:
Related to issue: https://github.com/sourcemeta/studio/issues/153
Following the existing CLI convention used for exit code 2, this PR updates those cannot-start paths to return exit code 3.


src/command_validate.cc
jviotti, line 1: Unintended changes?

src/command_lint.cc
jviotti, line 1: Unintended changes?

PR #654, opened 2026-02-10
Implement an `install` command

jviotti:

### DIFF
--- a/docs/install.markdown
+++ b/docs/install.markdown
@@ -1,36 +1,43 @@
 Installing Dependencies
-======================
+=======================
 
 ```sh
 jsonschema install
   [--force/-f] [--verbose/-v] [--debug/-g] [--json/-j]
 ```
 
-The `install` command fetches and installs external schema dependencies
-declared in the [`jsonschema.json`](./configuration.markdown) configuration
-file. It produces a `jsonschema.lock.json` lock file that tracks the hashes of
-installed dependencies, similar to how NPM uses `package-lock.json`.
+Many applications rely on consuming schemas authored and maintained by others.
+These schemas are typically published over HTTP(S) through registries like
+[schemas.sourcemeta.com](https://schemas.sourcemeta.com) (a free public
+registry) or self-hosted solutions like [Sourcemeta
+One](https://one.sourcemeta.com). While you could manually download these
+schemas or write a custom fetching script, doing so quickly gets complicated:
+schemas often reference other schemas that also need to be fetched, and the
+results need to be
+[bundled](https://json-schema.org/blog/posts/bundling-json-schema-compound-documents)
+to inline those references for local consumption. The `install` command solves
+this in a single step.
 
-The command discovers the nearest `jsonschema.json` by walking up from the
-current working directory. Dependencies are declared as a mapping of URIs to
-relative file paths:
+To get started, create a
+[`jsonschema.json`](./configuration.markdown) configuration file in your
+project and declare your dependencies as a mapping of schema URIs to local file
+paths. For example, to pull in the [JSON-RPC 2.0
+Response](https://schemas.sourcemeta.com/sourcemeta/std/v0/jsonrpc/v2.0/response)
+schema from the public registry:
 
 ```json
 {
   "dependencies": {
-    "https://example.com/schemas/user.json": "./vendor/user.json",
-    "https://example.com/schemas/product.json": "./vendor/product.json"
+    "https://schemas.sourcemeta.com/sourcemeta/std/v0/jsonrpc/v2.0/response": "./vendor/response.json"
   }
 }
 ```
 
-When you run `jsonschema install`, each dependency is fetched, bundled (to
-inline any remote references), and written to the specified path. The lock file
-`jsonschema.lock.json` is created or updated alongside `jsonschema.json` to
-record the hash of each installed dependency.
-
-On subsequent runs, the command only fetches dependencies that are missing,
-modified, or not yet tracked in the lock file.
+Then run `jsonschema install`. Each dependency is fetched, bundled, and written
+to the specified path. A `jsonschema.lock.json` lock file is created alongside
+`jsonschema.json` to record the hash of each installed dependency. On
+subsequent runs, only dependencies that are missing, modified, or not yet
+tracked in the lock file are fetched again.
 
 > [!TIP]
 > We recommend committing `jsonschema.lock.json` to version control (similar to


docs/install.markdown
```diff
@@ -0,0 +1,73 @@
...
+    "https://example.com/schemas/user.json": "./vendor/user.json",
+    "https://example.com/schemas/product.json": "./vendor/product.json"
+  }
+}
+```
+
+When you run `jsonschema install`, each dependency is fetched, bundled (to
+inline any remote references), and written to the specified path. The lock file
+`jsonschema.lock.json` is created or updated alongside `jsonschema.json` to
+record the SHA-256 hash of each installed dependency.
```
jviotti: Don't mention the specific hash algorithm, which might change in the future?

docs/install.markdown
```diff
@@ -0,0 +1,73 @@
...
+
+On subsequent runs, the command only fetches dependencies that are missing,
+modified, or not yet tracked in the lock file.
+
+> [!TIP]
+> We recommend committing `jsonschema.lock.json` to version control (similar to
+> `package-lock.json`) to enable reproducible installs across environments.
+
+> [!NOTE]
```
jviotti: Don't make it a note but some sort of warning. With a foot note that we are trying to fix this automatically in some other way?

docs/install.markdown
```diff
@@ -0,0 +1,73 @@
...
+jsonschema install --force
+```
+
+### Install with verbose output
+
+```sh
+jsonschema install --verbose
+```
+
+### No dependencies declared
```
jviotti: Remove this case altogether. No need to mention it

src/command_install.cc
```diff
@@ -0,0 +1,350 @@
...
+  std::string result{label};
+  result.append(14 - label.size(), ' ');
+  result += " : ";
+  return result;
+}
+
+auto read_json_file(const std::filesystem::path &path)
+    -> sourcemeta::core::JSON {
+  auto stream{sourcemeta::core::read_file(path)};
+  return sourcemeta::core::parse_json(stream);
```
jviotti: We have `sourcemeta::core::read_json` for this, which can make this entire helper unnecessary?

src/command_install.cc
```diff
@@ -0,0 +1,350 @@
...
+}
+
+auto atomic_write(const std::filesystem::path &path,
```
jviotti, line 31: Add a TODO note for elevating this to Core's IO module

src/command_install.cc
```diff
@@ -0,0 +1,350 @@
...
+  stream << "\n";
+  stream.close();
+  std::filesystem::rename(temporary_path, path);
+}
+
+auto http_fetch(const std::string &uri,
+                const sourcemeta::core::Options &options)
+    -> sourcemeta::core::JSON {
+  std::string fetch_url{uri};
+  if (fetch_url.find('?') != std::string::npos) {
```
jviotti: This whole thing feels very weak. Can we properly use `sourcemeta::core::URI` here? Hm, actually we don't allow setters for query in the URI module. Can you add a TODO comment for that? i.e. we should in theory be able to do this with the URI module

src/command_install.cc
```diff
@@ -0,0 +1,350 @@
...
+    fetch_url += "&bundle=1";
+  } else {
+    fetch_url += "?bundle=1";
+  }
+
+  cpr::Response response;
+  for (std::uint8_t attempt{1}; attempt <= HTTP_MAXIMUM_RETRIES; ++attempt) {
```
jviotti: This whole logic is duplicated in `src/resolver.h`. Maybe `src/resolver.h` can have this as another helper that both files use?

src/command_install.cc
```diff
@@ -0,0 +1,350 @@
...
+    }
+  }
+
+  if (response.status_code != 200) {
+    std::ostringstream error;
+    error << "HTTP " << response.status_code << "\n  at " << uri;
+    throw std::runtime_error(error.str());
+  }
+
+  return sourcemeta::core::parse_json(response.text);
```
jviotti: The duplicated logic for this in `src/resolver.h` accounts for YAML and the content type. This should do as well, and we should cover that with tests

src/command_install.cc
```diff
@@ -0,0 +1,350 @@
...
+    return;
+  }
+
+  static const char *type_names[] = {
+      "fetch/start",   "fetch/end",    "bundle/start", "bundle/end",
+      "write/start",   "write/end",    "verify/start", "verify/end",
+      "up-to-date",    "file-missing", "orphaned",     "mismatched",
+      "path-mismatch", "untracked",    "error"};
+  const auto type_index{static_cast<std::uint8_t>(event.type)};
+  std::cerr << "debug: " << type_names[type_index] << ": " << event.uri << " ("
```
jviotti, line 106: Let's add at least one test case with `--debug` then. Like a `_debug.sh` variant?

src/command_install.cc
```diff
@@ -0,0 +1,350 @@
...
+}
+
+} // namespace
+
+auto sourcemeta::jsonschema::install(const sourcemeta::core::Options &options)
+    -> void {
+  const auto configuration_path{
+      sourcemeta::blaze::Configuration::find(std::filesystem::current_path())};
+  if (!configuration_path.has_value()) {
+    throw std::runtime_error(
```
jviotti: Can we have a new error class in `src/error.h` for this case? That way when handling it, we can (when not `--json`) print a little footnote like we do for other errors linking to the docs in `docs/configuration.markdown` to learn more about this manifest file?

jviotti: Or actually link it to the URL of the install docs

src/command_install.cc
```diff
@@ -0,0 +1,350 @@
...
+    lock = sourcemeta::blaze::Configuration::Lock::from_json(
+        read_json_file(lock_path));
+  }
+
+  const auto fetch_mode{
+      options.contains("force")
+          ? sourcemeta::blaze::Configuration::FetchMode::All
+          : sourcemeta::blaze::Configuration::FetchMode::Missing};
+
+  const sourcemeta::blaze::Configuration::ReadCallback reader{
```
jviotti: Why assign it to a variable? Just use `configuration_reader` directly everywhere?

src/command_install.cc
```diff
@@ -0,0 +1,350 @@
...
+
+  const sourcemeta::blaze::Configuration::WriteCallback writer{
+      [](const std::filesystem::path &path,
+         const sourcemeta::core::JSON &document) -> void {
+        std::filesystem::create_directories(path.parent_path());
+        atomic_write(path, document);
+      }};
+
+  const sourcemeta::blaze::Configuration::FetchCallback fetcher{
+      [&options](std::string_view uri) -> sourcemeta::core::JSON {
```
jviotti: This is a bigger lambda. Move it as a standalone helper function.

src/command_install.cc
```diff
@@ -0,0 +1,350 @@
...
+          return read_json_file(parsed_uri.to_path());
+        }
+
+        if (scheme.has_value() &&
+            (scheme.value() == "https" || scheme.value() == "http")) {
+          return http_fetch(std::string{uri}, options);
+        }
+
+        std::ostringstream error;
+        error << "Could not resolve schema: " << uri;
```
jviotti: There is a specific error class for this case in `vendor/core`. And here we should throw it with a `FileError` wrapper pointing at the `jsonschema.json` file that had that info

src/command_install.cc
```diff
@@ -0,0 +1,350 @@
...
+        throw std::runtime_error(error.str());
+      }};
+
+  const sourcemeta::core::SchemaResolver install_resolver{
+      [&options, &configuration](std::string_view identifier)
```
jviotti, line 203: Also keep this lambda as a separate helper function in this class.

src/command_install.cc
```diff
@@ -0,0 +1,350 @@
...
+          -> std::optional<sourcemeta::core::JSON> {
+        const std::string string_identifier{identifier};
+
+        auto match{configuration.resolve.find(string_identifier)};
```
jviotti: I think a lot of this logic is duplicated in `src/resolver.h`. Find a way to re-use. Maybe `src/resolver.h` can expose some of this as helpers both use or something like that

src/command_install.cc
```diff
@@ -0,0 +1,350 @@
...
+              const auto path{absolute_uri.to_path()};
+              if (std::filesystem::exists(path)) {
+                return std::optional<sourcemeta::core::JSON>{
+                    read_json_file(path)};
+              }
+            }
+          }
+        }
+
+        for (const auto &[dependency_uri, dependency_path] :
```
jviotti: This logic now needs to apply for ALL other commands in `src/resolver.h`. Make sure that's the case and test this for other commands. A good candidate is `test/bundle`. i.e. if you install a dependency and then bundle something that depends on it, the bundle command should pick it up automatically based on `dependencies`

src/command_install.cc
```diff
@@ -0,0 +1,350 @@
...
+        }
+
+        sourcemeta::core::URI uri;
+        try {
+          uri = sourcemeta::core::URI{std::string{identifier}};
+        } catch (const sourcemeta::core::URIParseError &) {
+          return std::nullopt;
+        }
+
+        if (uri.is_file()) {
```
jviotti: This part seems also duplicated with the fetcher?

src/command_install.cc
```diff
@@ -0,0 +1,350 @@
...
+        emit_debug(options, event);
+
+        switch (event.type) {
+          case Type::FetchStart:
+            std::cerr << padded_label("Fetching") << event.uri << "\n";
+            break;
+          case Type::FetchEnd:
+            break;
+          case Type::BundleStart:
+            if (options.contains("verbose") || options.contains("debug")) {
```
jviotti: We have `LOG_VERBOSE` to replace these conditionals. Use that everywhere

src/command_install.cc
```diff
@@ -0,0 +1,350 @@
...
+      }};
+
+  configuration.fetch(lock, fetcher, install_resolver, reader, writer, on_event,
+                      fetch_mode);
+
+  if (had_error) {
+    throw sourcemeta::jsonschema::Fail{EXIT_FAILURE};
+  }
+
+  atomic_write(lock_path, lock.to_json());
```
jviotti, line 345: We always write a lock file, but NONE of the tests exercise the lock file. Every test should either assert that the lock exists or not and then assert its ENTIRE contents with `diff`

src/main.cc
```diff
@@ -122,6 +122,13 @@ Global Options:
 
        Decode a JSON document or JSONL dataset using JSON BinPack.
 
+   install [--force/-f]
+
+       Fetch and install external schema dependencies declared in
+       jsonschema.json. Uses the lock file jsonschema.lock.json to track
```
jviotti: No need to mention the lock file here. Just focus on the purpose of the command

test/install/fail_no_configuration.sh
jviotti, line 1: Here also assert that a `jsonschema.lock.json` was NOT created in the current working directory

test/install/pass_creates_parent_directories.sh
```diff
@@ -0,0 +1,37 @@
...
+"$1" install > "$TMP/output.txt" 2>&1
+
+cat << EOF > "$TMP/expected.txt"
+Fetching       : file://$(realpath "$TMP")/source/user.json
+Installed      : $(realpath "$TMP")/project/deep/nested/vendor/user.json
+EOF
+
+diff "$TMP/output.txt" "$TMP/expected.txt"
+
+test -f "$TMP/project/deep/nested/vendor/user.json"
```
jviotti: Never only check for file existence. Always assert the entire contents in every test

test/install/pass_multiple_dependencies.sh
```diff
@@ -0,0 +1,49 @@
...
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "type": "integer"
+}
+EOF
+
+cat << EOF > "$TMP/project/jsonschema.json"
+{
+  "dependencies": {
+    "file://$(realpath "$TMP")/source/alpha.json": "./vendor/alpha.json",
+    "file://$(realpath "$TMP")/source/beta.json": "./vendor/beta.json"
```
jviotti, line 30: I think we need a fail test where two dependencies point to the same path?

test/install/pass_resolver_dependency_on_disk.sh
```diff
@@ -0,0 +1,44 @@
...
+{
+  "\$schema": "file://$(realpath "$TMP")/source/metaschema.json",
+  "type": "object"
+}
+EOF
+
+cat << EOF > "$TMP/project/jsonschema.json"
+{
+  "dependencies": {
+    "file://$(realpath "$TMP")/source/metaschema.json": "./vendor/meta.json",
```
jviotti, line 34: Let's test what happens if the target file ALREADY exists by the time we do the first install?

test/install/pass_resolver_official_schema.sh
```diff
@@ -0,0 +1,36 @@
...
+TMP="$(mktemp -d)"
+clean() { rm -rf "$TMP"; }
+trap clean EXIT
+
+mkdir -p "$TMP/source" "$TMP/project"
+
+cat << 'EOF' > "$TMP/source/schema.json"
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "type": "object",
```
jviotti: As per the test name, should we make this schema actually do a `$ref` to an official meta-schema?

docs/install.markdown
jviotti: Fixed - removed the specific hash algorithm mention

docs/install.markdown
jviotti: Changed to WARNING with a footnote about working on automatic resolution

docs/install.markdown
jviotti: Removed this section

src/command_install.cc
jviotti: Replaced with `sourcemeta::core::read_json` and removed the helper

src/command_install.cc
jviotti: Added TODO comment

src/command_install.cc
jviotti: Added a TODO for using the URI module once it supports query setters

src/command_install.cc
jviotti: Extracted `http_fetch()` shared helper in resolver.h. Both command_install.cc and fallback_resolver now use it

src/command_install.cc
jviotti: The shared `http_fetch()` helper now handles YAML content type, matching the resolver.h behavior

src/command_install.cc
jviotti: Added `pass_debug_output.sh` test

src/command_install.cc
jviotti: Added `ConfigurationNotFoundError` class in error.h with catch handler that shows footnote linking to install docs

jviotti: Done - links to the install docs URL

src/command_install.cc
jviotti: Removed - now passing `configuration_reader` directly

src/command_install.cc
jviotti: Extracted to standalone `install_fetch()` helper function

src/command_install.cc
jviotti: Now throws `FileError<SchemaResolutionError>` pointing at the jsonschema.json path

src/command_install.cc
jviotti: Extracted to standalone `install_resolve()` helper function

src/command_install.cc
jviotti: Extracted `find_resolve_match()` shared helper in resolver.h. Both CustomResolver and install_resolve use it

src/command_install.cc
jviotti: Added dependency schema preloading in CustomResolver constructor. Installed dependencies are registered by their \$id and available to all commands. Added `bundle/pass_resolve_installed_dependency.sh` test

src/command_install.cc
jviotti: The install_resolve and install_fetch are now separate standalone functions. The file:// handling in install_resolve is for the bundling resolver (resolving \$refs) while install_fetch handles the top-level dependency fetching - these serve different purposes

src/command_install.cc
jviotti: Replaced all manual `options.contains("verbose") || options.contains("debug")` conditionals with `LOG_VERBOSE(options)`

src/command_install.cc
jviotti: Every test now asserts lock file existence and diffs its entire contents. Tests with no dependencies assert the lock file does NOT exist

src/main.cc
jviotti: Simplified - removed lock file mention

test/install/fail_no_configuration.sh
jviotti: Added assertion that lock file was NOT created

test/install/pass_creates_parent_directories.sh
jviotti: Now asserts entire schema file contents and full lock file with diff

test/install/pass_multiple_dependencies.sh
jviotti: This would require changes in the blaze library to detect duplicate paths. Skipping for now as it's beyond the scope of this PR

test/install/pass_resolver_dependency_on_disk.sh
jviotti: Added `pass_target_already_exists.sh` test that pre-creates the target with different contents and verifies install overwrites it correctly

test/install/pass_resolver_official_schema.sh
jviotti: Updated to have the schema `$ref` the official 2020-12 meta-schema

test/install/pass_multiple_dependencies.sh
jviotti: Correction: Added `fail_duplicate_destination.sh` test that verifies two dependencies pointing to the same destination path fails with exit code 1 and does not create a lock file. The test passes successfully.

src/resolver.h
```diff
@@ -138,6 +158,26 @@ class CustomResolver {
         }
       }
     }
+
+    if (this->configuration_.has_value()) {
+      for (const auto &[dependency_uri, dependency_path] :
```
jviotti, line 212: I think a problem here is that we don't register the schemas `dependency_uri` too. We should still be able to use it using THIS URI. Can we have a bundle test for that too?

test/bundle/pass_resolve_installed_dependency.sh
```diff
@@ -0,0 +1,58 @@
...
+  "dependencies": {
+    "file://$(realpath "$TMP")/source/types.json": "./vendor/types.json"
+  }
+}
+EOF
+
+cd "$TMP/project"
+"$1" install > /dev/null 2>&1
+
+# Now create a schema that references the installed dependency
```
jviotti: Let's avoid all comments in tests

test/install/fail_duplicate_destination.sh
```diff
@@ -0,0 +1,41 @@
...
+{
+  "dependencies": {
+    "file://$(realpath "$TMP")/source/a.json": "./vendor/schema.json",
+    "file://$(realpath "$TMP")/source/b.json": "./vendor/schema.json"
+  }
+}
+EOF
+
+cd "$TMP/project"
+"$1" install > "$TMP/output.txt" 2>&1 \
```
jviotti, line 38: We should fully assert on the output. Plus then re-run with `--json` like other tests to also confirm the error message as JSON

src/error.h
```diff
@@ -154,6 +154,13 @@ class UnknownCommandError : public std::runtime_error {
   std::string command_;
 };
 
+class ConfigurationNotFoundError : public std::runtime_error {
+public:
+  ConfigurationNotFoundError()
```
jviotti: Let's make this error also take the base directory from where we start the search (i.e. the current working directory in this case). But make it explicit and don't default it. Then expose that directory with a method in a way such that the error handling can print it, which I believe it can if its `.path()`

test/install/pass_creates_parent_directories.sh
jviotti, line 1: Add a variant of this test where we try to cross over a file and not a directory. i.e. imagine `./foo/bar` is a file but we try to install a dependency to `./foo/bar/baz.json`

test/install/pass_debug_output.sh
```diff
@@ -0,0 +1,45 @@
...
+    "file://$(realpath "$TMP")/source/user.json": "./vendor/user.json"
+  }
+}
+EOF
+
+cd "$TMP/project"
+"$1" install --debug > "$TMP/output.txt" 2>&1
+
+# Debug output should contain debug-prefixed lines
+grep -q "^debug: fetch/start:" "$TMP/output.txt"
```
jviotti: No `grep`. Let's be explicit on the ENTIRE output using `diff`

test/install/pass_debug_output.sh
```diff
@@ -0,0 +1,45 @@
...
+grep -q "^debug: write/start:" "$TMP/output.txt"
+grep -q "^debug: write/end:" "$TMP/output.txt"
+grep -q "^debug: verify/start:" "$TMP/output.txt"
+grep -q "^debug: verify/end:" "$TMP/output.txt"
+
+# Normal output should also be present
+grep -q "^Fetching" "$TMP/output.txt"
+grep -q "^Installed" "$TMP/output.txt"
+
+test -f "$TMP/project/vendor/user.json"
```
jviotti: Same here. Never only check for file existence. Check the entire contents with `diff` always

test/install/pass_dependency_as_metaschema.sh
```diff
@@ -0,0 +1,72 @@
...
+    "file://$(realpath "$TMP")/source/custom_meta.json": "./vendor/custom_meta.json",
+    "file://$(realpath "$TMP")/source/my_schema.json": "./vendor/my_schema.json"
+  }
+}
+EOF
+
+cd "$TMP/project"
+"$1" install > /dev/null 2>&1
+
+test -f "$TMP/project/vendor/custom_meta.json"
```
jviotti: Assert fully on the contents always. Never do `test -f`

test/install/pass_empty_dependencies.sh
```diff
@@ -0,0 +1,28 @@
...
+{
+  "dependencies": {}
+}
+EOF
+
+cd "$TMP/project"
+"$1" install > "$TMP/output.txt" 2>&1
+
+cat << 'EOF' > "$TMP/expected.txt"
+No dependencies to install
```
jviotti: I think this error is too basic. Not only we want to point out that no dependencies were found, but maybe the path of the `jsonschema.json`. Maybe:

```
No dependencies found
  at /absolute/path/to/jsonschema.json
```

test/install/pass_fetch_file_uri.sh
```diff
@@ -0,0 +1,65 @@
...
+cat << EOF > "$TMP/project/jsonschema.json"
+{
+  "dependencies": {
+    "file://$(realpath "$TMP")/source/schema.json": "./vendor/schema.json"
+  }
+}
+EOF
+
+cd "$TMP/project"
+"$1" install > /dev/null 2>&1
```
jviotti: Never pipe stuff to `/dev/null`. Always, in every single case, assert on the output of the commands using `diff`

test/install/pass_lock_file_format.sh
jviotti, line 1: I think we can remove this test. If all tests assert on the lock file, then we don;t need this one?

PR #655, opened 2026-02-11
Support adding new dependencies in `install`

jviotti:


src/command_install.cc
```diff
@@ -135,8 +135,65 @@ auto sourcemeta::jsonschema::install(const sourcemeta::core::Options &options)
...
+      }
+    } else {
+      configuration_path = std::filesystem::current_path() / "jsonschema.json";
+      add_configuration.absolute_path = std::filesystem::current_path();
+      add_configuration.base =
+          sourcemeta::core::URI::from_path(add_configuration.absolute_path)
+              .recompose();
+    }
+
+    const sourcemeta::core::URI uri{dependency_uri};
```
jviotti: TODO: Test for exceptions here?

test/install/pass_add_dependency_no_config.sh
```diff
@@ -0,0 +1,68 @@
...
+Fetching       : file://$(realpath "$TMP")/source/user.json
+Installed      : $(realpath "$TMP")/project/vendor/user.json
+EOF
+
+diff "$TMP/output.txt" "$TMP/expected.txt"
+
+cat << EOF > "$TMP/expected_config.json"
+{
+  "path": "$(realpath "$TMP")/project",
+  "baseUri": "file://$(realpath "$TMP")/project",
```
jviotti: This looks off. Why do we create a default path and baseURI? We should only have dependencies?

test/install/fail_add_duplicate_path.sh
```diff
@@ -0,0 +1,45 @@
...
+
+cd "$TMP/project"
+"$1" install "https://example.com/schema" "./vendor/user.json" \
+  > "$TMP/output.txt" 2>&1 && CODE="$?" || CODE="$?"
+test "$CODE" = "1"
+
+cat << EOF > "$TMP/expected.txt"
+error: Multiple dependencies cannot point to the same path
+  at file path $(realpath "$TMP")/project/jsonschema.json
+  at location "/dependencies/https:~1~1example.com~1schema"
```
jviotti: This looks off. The location is pointing at the pointer of the new dependency that will never be added. Instead we should be pointing at the existing location that has the same path?

PR #662, opened 2026-02-17
Support registering custom `lint` rules using `jsonschema.json`

jviotti:

### DIFF
--- a/src/command_lint.cc
+++ b/src/command_lint.cc
@@ -198,17 +198,16 @@ auto sourcemeta::jsonschema::lint(const sourcemeta::core::Options &options)
   }
 
   std::unordered_set<std::string> seen_configurations;
-  std::vector<std::pair<std::filesystem::path, bool>> input_paths;
+  std::vector<std::filesystem::path> input_paths;
   if (options.positional().empty()) {
-    input_paths.emplace_back(std::filesystem::current_path(), false);
+    input_paths.emplace_back(std::filesystem::current_path());
   } else {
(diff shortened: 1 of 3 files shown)


docs/lint.markdown
```diff
@@ -126,6 +126,23 @@ You can pass multiple custom rules:
 jsonschema lint --rule rule1.json --rule rule2.json path/to/my/schema.json
 ```
 
+Alternatively, you can declare custom rules in your
```
jviotti: This explanation is too long. Let's just say we can declare custom rules in `jsonschema.json` too and link to that without further examples or explanation.

test/lint/fail_lint_config_rule_missing_title.sh
```diff
@@ -0,0 +1,61 @@
...
+  "type": "string"
+}
+EOF
+
+cd "$TMP"
+"$1" lint "$TMP/schema.json" \
+  > "$TMP/output.txt" 2>&1 && CODE="$?" || CODE="$?"
+test "$CODE" = "1" || exit 1
+
+printf '%s\n' \
```
jviotti: No `printf` in any tests

test/lint/fail_lint_config_rule_missing_title.sh
```diff
@@ -0,0 +1,61 @@
...
+  "error: The schema rule title is missing or not a string" \
```
jviotti: This error is confusing. We already know the title is missing? Why do we say "or not a string"?

PR #861, opened 2026-09-11
Support --trace/-t in the test command

### LINKED ISSUE #307: Support passing `--trace` to `test` to automatically print traces (success or failure) on failed tests

magic-peach:
The `validate` and `metaschema` commands both support `--trace/-t` to print the evaluation trace on failure, but `test` had no equivalent, despite issue #307 asking for exactly this.

This adds a `--trace/-t` flag to the `test` command, wired into the same failure path that already prints the explanation for a failed test case, so a trace is emitted right after it. Passing `--trace` together with `--json` is rejected the same way `validate` already rejects that combination for itself.

Verified by adding two new end to end test cases (`fail_trace`, `fail_trace_json`) that check the actual output of the built binary, and by running the full `test` and `validate` suites locally: 130 and 267 tests respectively, all passing.

Fixes #307

<a href="https://cubic.dev/pr/sourcemeta/jsonschema/pull/861?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/command_test.cc
+++ b/src/command_test.cc
@@ -208,6 +208,7 @@ auto run_suite_as_text(const sourcemeta::core::Options &options,
         sourcemeta::jsonschema::default_dialect(options, configuration)};
     const auto &schema_resolver{sourcemeta::jsonschema::resolver(
         options, options.contains("http"), dialect, configuration)};
+    const auto trace{options.contains("trace")};
 
     auto test_suite{parse_test_suite(
         entry, schema_resolver, dialect,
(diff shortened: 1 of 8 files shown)


magic-peach:
Good catch on the double evaluation. It only applies to the failure path with --trace explicitly passed, which is already the slow, opt-in path for the schemas that failed. Doing it in a single pass would mean building a message-producing collector on top of TraceOutput's step-by-step callback instead of reusing SimpleOutput, which felt like a bigger change than this PR should carry. Happy to follow up separately if that tradeoff is not acceptable.

Also pushed a fix: the new --trace/-t flag changed the test command's help text, which broke the three help_command/help_option_long/help_option_short golden output tests in CI. Fixed and squashed into the same commit, full suite passes locally now.

jviotti (approved):
Oh cool! Well done!

PR #862, opened 2026-09-14
Report a nonexistent decode input the same way every other command does

magic-peach:
The decode command opened its input file with a raw std::ifstream guarded only by assert(!input_stream.fail()), the only command in the CLI still doing this instead of the shared read_file helper every other command (encode and 7 others) already uses.

In a build where assertions are compiled in, a nonexistent .binpack path crashes the process outright: Assertion failed: (!input_stream.fail()), function decode, file command_decode.cc, line 61, exit code 134. Confirmed this directly by reverting just the fix and running decode against a missing file.

This switches to read_file, which routes through the same IOFileNotFoundError handling every other command already relies on, turning the crash into the standard error: No such file or directory that decode, validate, and the rest of the CLI already give for this class of error.

Verified by adding test/decode/fail_enoent.clitest, matching the existing test/validate/fail_schema_enoent.clitest convention, and running the full test suite locally: 1136 tests, all passing.

<a href="https://cubic.dev/pr/sourcemeta/jsonschema/pull/862?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


jviotti (approved):
Good catch!

Issue #80, opened 2024-06-11
Support OpenAPI?


DavidBiesack:
99% of the JSON Schema I work with are embedded in an OpenAPI document (several documents with $ref etc.). This GitHub repo does not mention OpenAPI - can the CLI be used with JSON Schema embedded OpenAPI documents? If so, how?

jviotti:
Hi @DavidBiesack , thanks a lot for reaching out! This is indeed a very good point I was thinking about lately. 

One approach is to extend commands like `lint` to automatically recognise an OpenAPI specification and i.e. apply themselves to the schemas in such specification. However, this approach might not work very well with other commands like `bundle` or `test`, unless we come up with a way to say "I want to bundle/test *this* specific schema part of the OpenAPI specification".

What do you think? Any ideas would be more than welcome. Was there a specific command you were interested in using in the context of OpenAPI?

DavidBiesack:
@jviotti initially, I would like to use `lint` (with `--fix`).
Some OpenAPI context sensitive linting(with fix) would be nice - i.e. add `unevaluatedProperties: false` to schemas, but only where "safe" to do so (i.e. not schemas that are $referenced elsewhere for composition, but OK for those used in request or response bodies in the OpenAPI, would be good targets for such context-specific linting.)
Longer term, if refactoring is added to the CLI, that would be awesome ... notably, "extract these properties into a reusable schema, replacing with an `allOf`", rename schemas (automatically locate references and update them, Building schema cross-reference tracking and impact analysis would be nice.) Other types of refactoring tasks that are error prone when done manually would be great.
Other refactoring, like after adding a property, propagate updated `example` to other schemas, would be awesome.
(Ideally, the CLI would work with YAML source as well as JSON source)

jviotti:
Thanks a lot for the feedback and ideas. Makes sense. I'll aim to get linting support for OpenAPI + YAML some time soon.

jviotti:
General YAML support shipped in v5.0.0 (https://github.com/sourcemeta/jsonschema/releases/tag/v5.0.0). OpenAPI interoperability would be the next step!

jeremyfiel:
here is some inspiration from a post I made about using ajv to validate OAS schemas. It's mainly for OAS 3.0.x, which is the superset/subset of draft-04.  It would work with OAS3.1.x+ but Ajv uses a different import for later drafts. I don't have any logic to detect the $schema or the OAS version in this answer. 

[...]

PR #870, opened 2026-09-20
Cover linting schemas embedded in OpenAPI paths with tests

magic-peach:
The lint command's OpenAPI support only had test coverage for schemas under components/schemas. The primary use case from issue #80, schemas embedded in request and response bodies under paths, had zero coverage despite the underlying implementation already handling it correctly.

Added a check test that catches an enum_with_type violation in a response body schema, and a fix test that autofixes a redundant type alongside const in a request body schema while leaving the rest of the document untouched.

Tested: ctest -R "_paths" passes 2/2, full lint suite passes 206/206 with no regressions.

<a href="https://cubic.dev/pr/sourcemeta/jsonschema/pull/870?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


jviotti (commented):
Oh good catch! Wonderful

jviotti (approved):

